PR #46, opened 2026-05-16
feat: add lock screen plugin

ayangweb:
image


ayangweb:
@ggbond268 You can try rerunning this CI job — the error looks like it was caused by a network issue.

PR #133, opened 2026-06-09
fix(l10n): remove unnecessary line break in about description

YangFong:
image

## Summary by CodeRabbit

* **Documentation**
  * Updated Chinese (Simplified) localization for the About description to correct newline encoding, ensuring line breaks display correctly.


ggbond268:
Thanks for the contribution. I feel like a line break isn't necessary here. Let's just remove it.

ggbond268:
The text here doesn't need line breaks; just remove the \n. thanks

PR #139, opened 2026-06-13
macOS 27: survive the menu-bar rehosting + best-effort right-click revival

Goldenmonstew:
## Summary

Keeps MacTools usable on the **macOS 27 beta** (26A5353q), whose rehosted single-window menu bar broke third-party status-item interaction, and adds a best-effort right-click revival. Scoped to the app/host layer (`Sources/App`, `Sources/Core`) — file-disjoint from the Launchpad batch (separate PR).

> **Why now:** on the beta, clicking our menu bar icon stopped working through the usual channels; without this, beta users can't open the panels at all.

### Menu bar survival on macOS 27
- Survive the rehosting: geometry-first status-item click handling + stub-host bounce suppression (the beta forwards a synthetic `leftMouseDown` with event numbers all `0` and a degenerate backing-window frame, so click identity is carried by **screen location**, not event number).
- Adopt `NSStatusItemExpandedInterfaceSession` (pure runtime selector adoption, zero macOS 27 symbols) so `didBegin`/`didEnd` drive panel open/close on the new host; `≤26` keeps the old path unchanged.
- Fade menu-bar panels out on dismissal; read live keyboard state for the secondary-click judgment; tolerate backward timestamp skew and activation-driven dismissal in the click suppressor.

### Best-effort right-click revival + fallback
- Listen-only `.cgSessionEventTap` observes physical `rightMouseDown` at the session layer and opens the secondary panel only when one lands on our status item (pure observer, live geometry, fail-closed, self-healing; pure hit-test geometry headless-tested).
- On seeds where the status item's backing window frame is degenerate the tap fails closed; a macOS-27-gated settings notice steers users to **Option+left-click**, the reliable secondary path. Temporary bridge — removable once the host exposes a real secondary-click event/API.

### Other confirmed macOS 27 beta fixes
- ActivityBar charts build on the macOS 27 SDK (conditional `ChartContent`).
- Four confirmed breakages (status-item anchor, gamma load/restore, battery log, OCR).
- HideNotch mask pinned to the physical notch; MiddleClick pairing recovers when a converted down's up never arrives; MenuBarHidden host gate shape-validated; strip quarantine from local plugin installs after trust validation.
- Beta private-surface probes consolidated into `scripts/beta-probes`; impact audit + Feedback drafts in `docs/`.

## Testing
- `make build` (full app) green on the macOS 27 SDK; `MenuBarSecondaryClickTapTests` 21/21 green (hit-test geometry, fail-closed, coordinate flip).
[...]

PR #142, opened 2026-06-16
macOS 27: revive menu-bar right-click + fix beta breakages (macOS ≤26 byte-identical)

Goldenmonstew:
Replaces #139 (closed). That PR adapted the menu bar to the macOS 27 beta but leaked several behavior changes onto shipping macOS ≤26 — release users reported laggy left/right clicks. This is a clean redo: the **same verified macOS 27 solution, with every beta-only behavior gated to macOS 27 so macOS ≤26 stays byte-for-byte identical to today's shipping behavior.**

## Why

macOS 27 (beta 26A5353q) rehosts every menu bar status item into one shared window. For third-party items the backing window is a stub, only a synthesized `leftMouseUp` reaches the action (modifiers stripped), and right mouse events are not routed to any high-level channel. Left-click, right-click and the secondary panel all needed adapting — without changing how the app behaves on macOS ≤26.

## What's in it (3 commits)

1. **Survive and revive the menu bar on macOS 27** — adopt `NSStatusItemExpandedInterfaceSession` (runtime selectors), geometry-first clicks + stub-host bounce suppression, a listen-only CGEvent tap that revives the right-click, the Option+left / live-keyboard secondary channel, a reopen escape-hatch, and a fail-closed `MenuBarHidden` host gate. All gated to macOS 27.
2. **Repair plugin beta breakages** — gamma capacity query (`CGDisplayGammaTableCapacity`), HideNotch physical-notch mask, MiddleClick pairing recovery, OCR capture validation, battery log query, ActivityBar SDK build. Each ≤26-safe.
3. **Docs + beta-probe scripts**.

> The plugin-install quarantine fix that was bundled in an earlier revision is now split out into #143 — it is an all-OS Gatekeeper/quarantine fix, not macOS 27-specific, so it ships on its own.

## macOS ≤26 is byte-identical

Every macOS 27 behavior is behind `#available(macOS 27.0, *)` / `isMacOS27OrLater`. On macOS ≤26 the app keeps: the legacy action route (right-click already works there), instant panel close (no fade — this was the reported lag), right/Control-click as the only secondary gesture (no Option), window-identity hit testing, the original notch-mask formula `max(aux, menu bar)`, the original Settings copy, and AppKit's default reopen behavior. New ≤26 guard tests lock the Option and notch-height branches, and the whole change set was audited cluster-by-cluster for ≤26 leaks.

## Right-click on macOS 27 — verified working

The listen-only tap **revives real right-click** on the rehosted bar. The catch was a red herring worth calling out: the stub window frame reads degenerate *at launch*, but the host finishes placing the item a moment later, so by click time the frame is a real icon rect. The tap reads the frame **live on every event**, so an on-icon right-click hits and opens the secondary panel — confirmed on 26A5353q from the on-device diagnostic log (dozens of `hit=true` → `featurePanel` opens). It requires Accessibility permission (it's a CGEvent tap); Option+left and the live-keyboard channel are no-permission alternatives. The in-app notice and README point users at the Accessibility grant rather than implying right-click is unavailable.

## Testing

Full `xcodebuild … test` suite green. macOS 27 behaviors verified live on 26A5353q: left-click open/close, **right-click open via the tap** (confirmed in the diagnostic log), Option+left secondary panel, menu-bar survival across relaunch. macOS ≤26 paths are unchanged by construction (gated) and covered by the new guard tests.

# MacTools macOS 27 Beta Support PR Review

## Executive Summary
Strong macOS 27 host-adaptation approach with extensive new geometry/event-tap logic and broad test coverage. **Stability gap (critical):** multiple `required init?(coder:)` implementations still contain `fatalError()`, which can crash the host process if any of these views are instantiated via NSCoder (e.g., accidental storyboard/XIB usage).

## 1. SECURITY (✅ No clear critical backdoors/exfil in provided changes)
- **Event tap:** `MenuBarSecondaryClickTap` is listen-only and hit-tests before invoking callbacks; no evidence of swallowing/intercepting events for unauthorized behavior.
- **Clipboard/privacy:** Repo contains existing `NSPasteboard` usage across Translator/Clipboard-related plugins, but nothing in the provided macOS 27 menu-bar right-click adaptations indicates new clipboard exfil/storage.
- **LaunchAgents/Daemons:** No evidence from the scan results that this PR introduces LaunchAgent/LaunchDaemon persistence.

## 2. STABILITY & ROBUSTNESS (❌ Fix required: remove `fatalError()` from `init(coder:)`)
Found additional host-crash sites:

- `Plugins/Launchpad/Sources/LaunchpadDragGrid.swift`
  - `required init?(coder: NSCoder) { fatalError() }` (two occurrences: ~line 218 and ~line 1318)
- `Sources/App/FeatureManagementTableView.swift`
  - `required init?(coder: NSCoder)` with `fatalError("init(coder:) has not been implemented")`
- `Plugins/MenuBarHidden/Sources/MenuBarHiddenComponentView.swift`
  - `fatalError("init(coder:) has not been implemented")`
- `Plugins/Calendar/Tests/CalendarEventServiceOpeningTests.swift`
  - `required init?(coder: NSCoder) { fatalError(...) }` (test code; should not be crash-path for the app)

### Actionable fix (committable)
Replace each production `fatalError()` in `required init?(coder:)` with a non-crashing initializer, e.g.:
```swift
required init?(coder: NSCoder) {
    super.init(frame: .zero)
}
```
(or the appropriate `super.init(...)` for the specific view subclass). For any test-only stubs, keep them non-fatal as well.

## 3. PERFORMANCE (✅ Likely OK)
- Secondary click recovery is **listen-only** and gated by Accessibility trust.
- Host compatibility gating and state-machine logic avoids repeated churn; no obvious infinite loops/timer leaks visible from the provided summary.

## 4. CODE CONVENTIONS & STYLING (✅ Consistent)
- Strong macOS 27 gating (`#available(macOS 27.0, *)` / runtime checks) and cluster-by-cluster confinement.
- SwiftUI `Chart` structural refactor to avoid macOS 27 SDK conformance pitfalls is appropriate.

## Recommended PR status
**Request changes** to remove remaining `fatalError()` usages in `required init?(coder:)` across the files above. Once replaced with safe initializers, the PR should be merge-ready.


Goldenmonstew:
@ggbond268 — ready whenever you have a moment 🙏

**What this is:** the clean, ≤26-safe redo of the macOS 27 menu-bar adaptation (replaces the closed #139, which had regressed click latency on shipping macOS). Every macOS 27-only behavior is gated behind `#available(macOS 27.0, *)` / `isMacOS27OrLater`, so **macOS ≤26 stays byte-identical to `main`** (with guard tests locking the gated branches) — the earlier "laggy click" regression cannot recur there.

**Status:** CI green, mergeable (just rebased onto `f33fc42`). On macOS 27, right-click on the icon is revived via a listen-only accessibility event tap (verified on device; Option+left is the no-permission fallback).

It's been green for a few days and keeps needing a rebase each time `main` advances — happy to address any review comments. The plugin-install quarantine fix is split out into #143 (all-OS, not beta-specific) in case you'd rather take that one separately.

Thanks!

ggbond268:
It seems to be a bug in macOS 27 beta, waiting for Apple to fix it. First, make a transition plan for a combination key to expand the right-click menu. Don't implement too heavy auxiliary function recognition.

https://www.reddit.com/r/MacOSBeta/comments/1u0td39/right_clicking_menu_bar_icons_not_doing_anything/
https://developer.apple.com/documentation/AppKit/NSStatusItem

PR #149, opened 2026-06-26
Finder context menu (FinderSync): copy paths, new file, open in Terminal, open with apps

Goldenmonstew:
## What

Adds a Finder right-click (context menu) enhancement, built as a system **FinderSync extension** — separate from the in-process plugin model, so it's unaffected by the macOS 27 menu-bar rehosting.

Actions, each toggleable in **Settings → 访达菜单**:
- **复制路径** ▸ absolute / shell-escaped / file name / `file://` URL
- **新建文件** ▸ `.txt` / `.md` / `.json`
- **在终端打开**
- **用应用打开** ▸ a user-configured app list, with an optional per-app extension filter

## How

- **`FinderSyncExtension/`** — a `FIFinderSync` app-extension target, system-loaded into Finder. Copy actions run in-extension (the pasteboard needs no file access); file-creating / app-launching actions are forwarded to the non-sandboxed host app over a `mactools://` URL scheme, keeping the extension at minimal sandbox entitlements.
- **Sharing config across the sandbox boundary** — the non-sandboxed host can't write the app-group container, and `UserDefaults(suiteName:)` doesn't bridge the two cfprefsd domains (both measured to fail). Resolved by writing JSON to `~/Library/Application Support/MacTools/` and reading it from the extension through a read-only `temporary-exception.files.home-relative-path` entitlement; both sides resolve the real home via `getpwuid` (a sandboxed process's `NSHomeDirectory()` points at its container).
- **Dynamic app identity** — the "open with" action carries the chosen app in the menu item *title*: FinderSync preserves the displayed title but strips `representedObject` and doesn't reliably carry `tag`.
- **Security** — the host's `mactools://` handler allow-lists new-file extensions (blocks path traversal via the `ext` param) and validates open-with requests (app must be an existing `.app`, at least one file must exist).

## Tests

19 unit tests:
- `FinderMenuConfigStore` — save/load round-trip, missing-file & corrupt-data fallback, intermediate-dir creation
- `FinderContextMenuRequestHandler` — ext allow-list (incl. traversal), `nextAvailableURL` collision, `parseOpenWithRequest` validation
- `FinderContextMenuLogic` — shell escaping
- `OpenWithApp.matches` — extension filtering

Full flow manually verified on macOS 27: all menu items render, config sharing works end-to-end (host writes → sandboxed extension reads), and all four actions execute (including click-through of the open-with submenu launching the chosen app).

## Notes

New user-facing entitlement on the extension: a **read-only** `temporary-exception.files.home-relative-path` scoped to `/Library/Application Support/MacTools/`. MacTools ships via Developer ID + notarization (not the App Store), where this is acceptable; it is read-only (the earlier write path was already removed in favor of URL-scheme forwarding).

Added a configurable Finder right-click context menu via a dedicated `FinderSyncExtension` (`FIFinderSync`), kept separate from the in-process plugin model.

### Highlights
- New settings entry: **Settings → 访达菜单**
- **Copy path** submenu: absolute path, shell-escaped path, file name, and `file://` URL
- **New file** submenu: `.txt`, `.md`, `.json`
- **Open in Terminal**
- **Open With** submenu: user-configured app list with optional per-app extension filtering

### Architecture
- `FinderSyncExtension/` implements `FinderContextMenuProvider`:
  - Builds the menu and performs in-extension copy actions via `NSPasteboard` (write-only).
  - For file creation and app launching, forwards requests to the host using a `mactools://` URL so the extension remains minimally sandboxed.
  - Selects the “Open With” target app using the menu item title because FinderSync does not reliably preserve `representedObject`/`tag`.
- Host side (`FinderContextMenuRequestHandler`) handles `mactools://`:
  - Validates allowed new-file extensions (`txt`, `md`, `json`) and prevents traversal/out-of-directory writes when creating non-colliding filenames.
  - Validates “open with” requests: app must be a real `.app` bundle; referenced files must exist.
- Shared configuration:
  - JSON stored at `~/Library/Application Support/MacTools/finder-menu.json`, with sandbox-safe resolution using `getpwuid/getuid` in `FinderMenuConfigStore`.
  - FinderSync entitlements grant read-only, home-relative-path access to the config directory.

### Tests
- Added 19 XCTest cases covering:
  - config persistence/loading defaults
  - request parsing/validation (including traversal/collision defenses)
  - path shell escaping
  - open-with app matching semantics (including extension-filter edge cases)

### Security / Stability / Performance Notes (Actionable)
- **Security posture:** No evidence of backdoor-like behaviors; actions are constrained by allow-lists/validation, and clipboard usage is limited to write operations for copy actions.
- **Performance risk (important):** `FinderContextMenuProvider.menu(for:)` calls `FinderMenuConfigStore.load()` synchronously (`Data(contentsOf:)` + JSON decode) on Finder’s menu callback. Consider caching the config in-memory and reloading asynchronously (or only when stale) so menu rendering never blocks on disk I/O.
  - Suggested fix: load once (or keep a cached value) and update via a background refresh; `menu(for:)` should return the cached config immediately.
- **UI correctness bug:** `addOpenWithMenu` currently adds each matched app menu item **twice** (duplicate `submenu.addItem(...)` lines). Remove the duplicate insertion to avoid duplicated entries in “Open With”.

### DIFF
--- a/Configs/MacTools-Info.plist
+++ b/Configs/MacTools-Info.plist
@@ -37,6 +37,8 @@
 	</array>
 	<key>LSUIElement</key>
 	<true/>
+	<key>MTAppGroupIdentifier</key>
+	<string>group.$(BUNDLE_IDENTIFIER_PREFIX).mactools</string>
 	<key>MTPluginCatalogPublicKey</key>
 	<string>$(PLUGIN_CATALOG_PUBLIC_KEY)</string>
 	<key>NSAppleEventsUsageDescription</key>
--- a/Configs/MacTools-Info.plist
+++ b/Configs/MacTools-Info.plist
@@ -24,6 +24,17 @@
 	<string>public.app-category.utilities</string>
 	<key>LSMinimumSystemVersion</key>
 	<string>$(MACOSX_DEPLOYMENT_TARGET)</string>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleURLName</key>
+			<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>mactools</string>
+			</array>
+		</dict>
+	</array>
 	<key>LSUIElement</key>
 	<true/>
 	<key>MTPluginCatalogPublicKey</key>
--- a/Configs/MacTools-Info.plist
+++ b/Configs/MacTools-Info.plist
@@ -37,8 +37,6 @@
 	</array>
 	<key>LSUIElement</key>
 	<true/>
-	<key>MTAppGroupIdentifier</key>
-	<string>group.$(BUNDLE_IDENTIFIER_PREFIX).mactools</string>
 	<key>MTPluginCatalogPublicKey</key>
 	<string>$(PLUGIN_CATALOG_PUBLIC_KEY)</string>
 	<key>NSAppleEventsUsageDescription</key>
--- a/Configs/MacTools.entitlements
+++ b/Configs/MacTools.entitlements
@@ -6,5 +6,9 @@
 	<true/>
 	<key>com.apple.security.personal-information.calendars</key>
 	<true/>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.$(BUNDLE_IDENTIFIER_PREFIX).mactools</string>
+	</array>
 </dict>
 </plist>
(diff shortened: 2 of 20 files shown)


Goldenmonstew:
Superseded by https://github.com/ggbond268/MacTools/pull/150 — rebuilt on the merged RightClick architecture (Plugins/RightClick + Sources/Core/RightClick) instead of the parallel FinderSync extension this PR introduced. Same features (shell-escaped/file:// copy, new file, open in Terminal, open with a configurable app list) plus per-item toggles, now integrated into the existing RightClick plugin.

PR #150, opened 2026-06-29
Extend Finder right-click menu: new file, open in Terminal, open with apps, configurable

Goldenmonstew:
## What

Extends the existing **RightClick** Finder tools (new folder + copy name / absolute / relative path, added in `d3072a5`) with the actions and configurability from #149, integrated into the same `Plugins/RightClick` + `Sources/Core/RightClick` architecture.

Added, each toggleable in the plugin's settings:
- **Copy**: shell-escaped path (for pasting as terminal arguments) and `file://` URL
- **New file**: `.txt` / `.md` / `.json`
- **Open in Terminal**
- **Open with**: a user-configured app list, with an optional per-app extension filter
- A **settings page** (plugin `PluginConfiguration`) to toggle every menu item and manage the open-with apps

## How

Follows the existing RightClick patterns (`ActionID` + `openHostURL` → `RightClickURLRouter` → `RightClickFileActionService`):
- `RightClickPathFormatter` gains `joinedShellEscapedPaths` / `joinedFileURLs`; `RightClickFileActionService` gains `createFile` (allow-listed types); the router gains `/new-file`, `/open-terminal`, `/open-with`.
- **Shared config across the sandbox boundary** (new `RightClickConfiguration` + store): the non-sandboxed host can't write the app-group container, and `UserDefaults(suiteName:)` doesn't bridge the two cfprefsd domains (both measured to fail). So the config is a JSON file under the real `~/Library/Application Support/MacTools/`, written by the host/plugin and read by the sandboxed extension through a **read-only** `temporary-exception.files.home-relative-path` entitlement; both sides resolve the real home via `getpwuid` (a sandboxed process's `NSHomeDirectory()` points at its container).
- The "open with" item carries the chosen app by **title** (the app name): FinderSync preserves the displayed title but can strip `representedObject` — the existing title fallback in `RightClickFinderSync` covers it.
- The extension reads the config on every menu build to honor the toggles.

## Security

- New-file extension allow list (`txt` / `md` / `json`) also blocks path traversal through the forwarded `ext` parameter.
- Open-with validates the app (must be an existing `.app`) and files (at least one must exist) in the host handler.

## Tests

23 unit tests: shell-escape / file:// formatting, new-file allow list + creation, config round-trip / missing-key tolerance / corrupt-data fallback, open-with extension matching + request parsing.

## Notes

- **Supersedes #149**, which added the same feature as a parallel FinderSync extension *before* the RightClick tools landed in `main`. Closing that in favor of this, which builds on the merged RightClick architecture instead of duplicating it.
- New extension entitlement: a **read-only** `temporary-exception.files.home-relative-path` scoped to `/Library/Application Support/MacTools/`. MacTools ships via Developer ID + notarization (not the App Store), where this is acceptable; it is read-only.

Expanded Finder right-click support with configuration-driven actions to create new files (`.txt`/`.md`/`.json`), open selections in Terminal, open selections with a user-configured “open with” app list (optionally filtered by extension), and copy multiple path formats (file name, absolute/relative paths, shell-escaped paths, and `file://` URLs).

Added shared JSON-backed configuration persisted at `~/Library/Application Support/MacTools/right-click-menu.json`, plus a SwiftUI settings page to toggle each menu item and manage the open-with apps list (including extension filter input normalization). Extended the host URL router and file-action service with validated routing/requests (new-file directory + allowlisted extension, open-with app-bundle + existing files only, including logging for async open failures) and safe filename/file creation behavior.

Updated FinderSync menu construction to respect runtime configuration and route new actions, and added/extended tests covering config persistence/decoding fallback, open-with URL parsing and matching, shell/path formatting, allowlisted new-file creation, and extension filter behavior.


Goldenmonstew:
Addressed all 8 CodeRabbit comments across 4 fix commits:

- Strip a leading `.` from extension filters (pathExtension is dotless)
- Exclusive file create (`Data.write(.withoutOverwriting)`) to close the TOCTOU clobber gap
- Log `NSWorkspace.open` failures in both the open-in-terminal and open-with paths
- Constrain `/open-with` to apps present in the saved config (the URL scheme can be invoked by any process)
- Forward only the files the chosen app's extension filter matches (mixed selections no longer send excluded types)
- Carry the stable `appPath` (not the display name) in `representedObject`; disambiguate duplicate titles for the strip-safe fallback
- Two-item assertion to lock down the multi-URL `file://` clipboard format
- Scope the read-only file exception to `right-click-menu.json` instead of the whole directory (least privilege)

Also verified end-to-end on macOS 27 that the sandboxed extension reads the config through the file-scoped exception (toggling `openInTerminal` off removes the menu item). 23 unit tests pass.

PR #186, opened 2026-07-13
Redesign plugin layouts, visibility, and Marketplace lifecycle

xcv58:
## Summary

- Reframes the Plugins workspace around three clear jobs: **Dashboard** layout, **Feature Panel** layout, and **Marketplace** lifecycle management.
- Gives Dashboard and Feature Panel independent show/hide controls and ordering. Eye buttons use the same visual language as plugin settings, while hidden plugins remain in a separate non-draggable section and return to their prior position when shown again.
- Keeps Marketplace responsible for discovery, installation, updates, and uninstallation. Layout rows also expose contextual lifecycle actions on right-click, so cleanup does not require repeated navigation between pages.
- Makes plugin icons a direct settings shortcut when that plugin has a settings page.
- Adds a Settings-session option to pause repeated uninstall confirmations while preserving confirmation by default.
- Stabilizes repeated drag reordering by keeping draggable rows AppKit-native and deferring drop-state updates outside the table-view layout cycle.
- Preserves existing orders and visibility during upgrade, export, and import, including backups created by older app versions and conservative projections for downgrades.
- Refreshes user-facing copy in every supported app language and expands regression coverage for layout, migration, backup, lifecycle, drag, and localization behavior.

> [!IMPORTANT]
> Plugins previously disabled from the former Installed page now load after upgrade and may resume background behavior. They remain hidden on every supported surface until shown. Users who do not want a plugin running should uninstall it.

## Why

The previous Installed page mixed plugin lifecycle, visibility, and ordering. This revision separates those responsibilities:

- **Marketplace** controls whether a plugin is installed.
- **Dashboard** controls whether and where an installed Dashboard-capable plugin appears.
- **Feature Panel** controls whether and where an installed Feature Panel-capable plugin appears.
- **Plugin settings** contain only plugin-specific configuration.

This preserves the familiar visibility behavior of the former Installed checkbox without introducing a second global enabled/disabled state. A plugin hidden from both surfaces remains installed and may continue background work; uninstalling it from Marketplace or a layout row is the lifecycle action that removes it.

### Dashboard

Dashboard shows visible plugins in drag order and keeps Dashboard-capable hidden plugins in a clearly separated section. The eye control changes only Dashboard visibility; clicking a plugin icon opens its settings when available.

Dashboard layout

### Feature Panel — top

Feature Panel uses the same eye-and-drag interaction while maintaining order and visibility independently from Dashboard.

Feature Panel layout top

### Feature Panel — bottom

Hidden Feature Panel plugins appear in a separate non-draggable section. Showing one restores its prior position in the active ordered list.

Feature Panel layout bottom

### Marketplace

Marketplace is the lifecycle surface for discovering, installing, updating, and uninstalling plugins. Installed cards describe the surfaces each plugin supports.

Marketplace

### Plugin settings

Plugin settings remain scoped to plugin-specific configuration. System Status uses the same eye style for component and menu-bar metric visibility, providing a consistent show/hide convention throughout the app.

System Status settings and sidebar

## Upgrade, downgrade, and backup compatibility

- The previous shared plugin order seeds both Dashboard and Feature Panel during upgrade; subsequent ordering remains independent per surface.
- The legacy global hidden state maps to every supported surface for each available plugin.
- Temporarily unavailable plugin IDs remain staged until their capabilities can be resolved, avoiding accidental visibility resets.
- The old package-level migration marker is cleared only after the new display state is durably stored. If an unknown future schema is preserved, the marker remains until an explicit edit succeeds; relaunch and uninstall/reinstall regressions cover this handoff.
- New backups store independent Dashboard and Feature Panel orders and hidden sets.
- Legacy order and hidden fields remain populated as a compatibility projection, and older backups continue to import.
- Unknown newer preference payloads survive passive reads and lazy migration; an explicit user layout edit intentionally replaces them with the current schema.
- PluginKit v3 keeps its previously published visibility-lifecycle protocol symbol so already-installed plugin binaries remain loadable.
- If a plugin is uninstalled while a single or bulk update is resolving, uninstall remains authoritative and the late update is discarded.
- Older app versions can read the retained global order and visibility projection, but independent per-surface visibility collapses conservatively and the former disabled/not-loaded lifecycle state cannot be reconstructed after upgrading.

Migration is a small synchronous preferences transformation with no plugin installation or long-running data conversion, so no upgrade progress UI is needed.

## Release coordination

- Package code changed for Fan Control, Hide Menu Bar Icons, and System Status. Their manifest versions intentionally remain unchanged in this feature PR; the standard plugin release preparation detects these package changes and applies patch bumps.
- Publish that plugin batch before the corresponding app release. MacTools checks installed plugins automatically once per app version, so publishing the app first could defer these package updates until a later app version or a manual Marketplace update.

## Validation

- Latest `origin/main` is the PR merge base: **0 commits behind**
- `make build`
- Full MacTools test suite: **1,057 tests passed, 0 failures**
- GitHub **Build and test** workflow passed on the final commit
- Focused display migration, relaunch, backup, PluginKit compatibility, package lifecycle, drag, and localization suites
- Regression coverage includes future schemas with legacy-compatible keys, durable legacy-marker acknowledgement, uninstall/reinstall, and single/bulk update-uninstall interleavings
- Binary symbol inspection confirms the PluginKit v3 compatibility protocol remains exported
- App and plugin changelog validation
- `git diff --check`
- Manual local Debug-build verification of Dashboard and Feature Panel visibility/reordering, Marketplace lifecycle actions, uninstall confirmation suppression, plugin-settings navigation, and System Status metric visibility


xcv58:
@ggbond268 This PR is still draft, but I’d love your product/design opinion before marking it ready for review.

The main direction is:

- Installed focuses on global plugin enable/disable and supported surfaces.
- Dashboard and Feature Panel become the layout/order editors.
- The rightmost checkbox consistently means global enable/disable.
- Disabled plugins are shown in a separate bottom section, but their saved layout position is preserved and restored when re-enabled.
- Plugin settings pages also expose the global enabled switch.
- Preference export/import now preserves separate Dashboard and Feature Panel orders, while older backup files remain backward compatible.

Does this separation of responsibilities feel right to you? The main product question I’d like feedback on is whether Dashboard/Feature Panel should show disabled plugins in a separate section while preserving their previous position behind the scenes.

xcv58:
Addressed the review feedback in fe3750d:\n\n- Fixed the enablement checkbox accessibility label/help so VoiceOver reads the plugin name as the control label and the global enable/disable action as help.\n- Made display preference fallback downgrade-safe: unknown future schemas and unreadable payloads now fall back in memory without deleting or overwriting the stored payload during read-time lazy migration.\n- Added an in-memory cache for decoded display preferences.\n- Clarified the settings-page enabled-state fallback and documented the legacy/surface ordering paths.\n- Added regression coverage for future-version and corrupt display preference payloads.\n\nValidation:\n- PluginDisplayPreferencesStoreTests\n- FeatureManagementTableViewTests\n- PluginHostComponentSupportTests\n- PreferencesBackupTests\n- make build

ggbond268:
> This PR is still draft, but I’d love your product/design opinion before marking it ready for review.
> 
> The main direction is:
> 
> Installed focuses on global plugin enable/disable and supported surfaces.
> Dashboard and Feature Panel become the layout/order editors.
> The rightmost checkbox consistently means global enable/disable.
> Disabled plugins are shown in a separate bottom section, but their saved layout position is preserved and restored when re-enabled.
> Plugin settings pages also expose the global enabled switch.
> Preference export/import now preserves separate Dashboard and Feature Panel orders, while older backup files remain backward compatible.
> Does this separation of responsibilities feel right to you? The main product question I’d like feedback on is whether Dashboard/Feature Panel should show disabled plugins in a separate section while preserving their previous position behind the scenes.

Thanks for sharing the draft! Here are my thoughts:

1. Separating layout/order from visibility for the Dashboard and Feature Panel sounds like a good direction. One thing to watch out for is that some plugins only support one surface. If a plugin doesn’t support the Dashboard or the Feature Panel, it probably shouldn’t be shown there at all.

2. I’m not convinced the global **Disable** option in the **Installed** page is necessary. A proper **Disable** feature would need to stop all background activity for the plugin, which adds quite a bit of complexity. Since installing and uninstalling plugins is lightweight, I’d rather just uninstall plugins I don’t need. That also saves disk space. So I’d consider replacing **Disable** with **Uninstall** instead.  Going a step further, I even think the Installed page could potentially be removed, with install/uninstall handled directly from the plugin marketplace using sorting and filtering.

xcv58:
> Thanks for sharing the draft! Here are my thoughts:
> 
> 1. Separating layout/order from visibility for the Dashboard and Feature Panel sounds like a good direction. One thing to watch out for is that some plugins only support one surface. If a plugin doesn’t support the Dashboard or the Feature Panel, it probably shouldn’t be shown there at all.
> 2. I’m not convinced the global **Disable** option in the **Installed** page is necessary. A proper **Disable** feature would need to stop all background activity for the plugin, which adds quite a bit of complexity. Since installing and uninstalling plugins is lightweight, I’d rather just uninstall plugins I don’t need. That also saves disk space. So I’d consider replacing **Disable** with **Uninstall** instead.  Going a step further, I even think the Installed page could potentially be removed, with install/uninstall handled directly from the plugin marketplace using sorting and filtering.

Great, thank you so much. I will do more iteration and ping you once the new interface is ready.

xcv58:
> Thanks for sharing the draft! Here are my thoughts:
> 
> 1. Separating layout/order from visibility for the Dashboard and Feature Panel sounds like a good direction. One thing to watch out for is that some plugins only support one surface. If a plugin doesn’t support the Dashboard or the Feature Panel, it probably shouldn’t be shown there at all.
> 2. I’m not convinced the global **Disable** option in the **Installed** page is necessary. A proper **Disable** feature would need to stop all background activity for the plugin, which adds quite a bit of complexity. Since installing and uninstalling plugins is lightweight, I’d rather just uninstall plugins I don’t need. That also saves disk space. So I’d consider replacing **Disable** with **Uninstall** instead.  Going a step further, I even think the Installed page could potentially be removed, with install/uninstall handled directly from the plugin marketplace using sorting and filtering.

@ggbond268 Thanks again for the direction — I’ve reworked the draft around it.
 - Removed the Installed page and the global enable/disable model.
 - Dashboard and Feature Panel are now order-only views. Each shows only installed plugins that support that specific surface, so unsupported plugins do not appear there.
 - Marketplace now owns plugin lifecycle: discovery, installation, updates, and uninstallation. It also shows each plugin’s supported surfaces.
 - To avoid forcing users to leave a layout while cleaning up plugins, layout rows offer contextual Uninstall and View in Marketplace actions. These are shortcuts to the same Marketplace lifecycle, not a separate state model.
 - Existing Dashboard/Feature Panel orders and older preference backups remain compatible through migration.

I also refreshed the PR screenshots and description to reflect this final model. When you have a moment, I’d really appreciate one more product/design check before I mark the PR ready for review.

ggbond268:
> > Thanks for sharing the draft! Here are my thoughts:
> 
> > 
> 
> > 1. Separating layout/order from visibility for the Dashboard and Feature Panel sounds like a good direction. One thing to watch out for is that some plugins only support one surface. If a plugin doesn’t support the Dashboard or the Feature Panel, it probably shouldn’t be shown there at all.
> 
> > 2. I’m not convinced the global **Disable** option in the **Installed** page is necessary. A proper **Disable** feature would need to stop all background activity for the plugin, which adds quite a bit of complexity. Since installing and uninstalling plugins is lightweight, I’d rather just uninstall plugins I don’t need. That also saves disk space. So I’d consider replacing **Disable** with **Uninstall** instead.  Going a step further, I even think the Installed page could potentially be removed, with install/uninstall handled directly from the plugin marketplace using sorting and filtering.
> 
> 
> 
> @ggbond268 Thanks again for the direction — I’ve reworked the draft around it.
> 
>  - Removed the Installed page and the global enable/disable model.
> 
>  - Dashboard and Feature Panel are now order-only views. Each shows only installed plugins that support that specific surface, so unsupported plugins do not appear there.
> 
>  - Marketplace now owns plugin lifecycle: discovery, installation, updates, and uninstallation. It also shows each plugin’s supported surfaces.
> 
>  - To avoid forcing users to leave a layout while cleaning up plugins, layout rows offer contextual Uninstall and View in Marketplace actions. These are shortcuts to the same Marketplace lifecycle, not a separate state model.
> 
>  - Existing Dashboard/Feature Panel orders and older preference backups remain compatible through migration.
> 
> 
> 
> I also refreshed the PR screenshots and description to reflect this final model. When you have a moment, I’d really appreciate one more product/design check before I mark the PR ready for review.

Hi! The new UI looks great. I think I may not have explained my idea clearly.

The Dashboard and Feature Panel should still have a show/hide capability (for example, using an eye icon) to control whether a plugin appears in that panel. This would be equivalent to the checkbox in the previous Installed page.
With this approach, old preferences backups should remain compatible as well.

I may have misunderstood your earlier proposal.

xcv58:
> The Dashboard and Feature Panel should still have a show/hide capability (for example, using an eye icon) to control whether a plugin appears in that panel. This would be equivalent to the checkbox in the previous Installed page. With this approach, old preferences backups should remain compatible as well.

No worries, this makes sense to me. I will work on this and share another version in a few days.

xcv58:
@ggbond268 Thanks again for clarifying the intended visibility model. I’ve updated the draft to match it:
Dashboard and Feature Panel now each have independent eye-style show/hide controls and drag ordering.

- Plugins appear only on surfaces they support.
- Hidden plugins remain installed, are shown in a separate non-draggable section, and return to their previous position when shown again.
- Marketplace remains responsible for installation, updates, and uninstallation—there is no separate global enabled/disabled state.
- The former Installed checkbox state migrates to the supported surfaces. Older backups still import, while new backups retain a compatibility projection for older app versions.
- I also stabilized repeated drag reordering and expanded the migration, backup, lifecycle, compatibility, and UI regression coverage. The branch is current with main; CI and the full local suite pass with 1,058 tests.

When you have a moment, could you please take another product/design look—especially at the eye controls and the separation between per-surface visibility and Marketplace lifecycle? If the direction now matches what you had in mind, I’ll mark the PR ready for formal review.

ggbond268:
> When you have a moment, could you please take another product/design look—especially at the eye controls and the separation between per-surface visibility and Marketplace lifecycle? If the direction now matches what you had in mind, I’ll mark the PR ready for formal review.

From a product and design perspective, everything looks good to me now. Feel free to mark the PR as ready for review. 👍

PR #189, opened 2026-07-14
Add Sidecar plugin

xcv58:
## Summary

- Add a native `sidecar` MacTools plugin with runtime SidecarCore loading, discovery, direct connect/disconnect actions, wired-only connection policy, and clear unsupported-system handling.
- When exactly one display is reliably reported connected, selecting another display or using its explicit Connect shortcut now performs a clear, callback-driven switch: disconnect the current display first, then connect the target. Failures state which display remains connected or was disconnected.
- Add connected-status presentation, saved per-display settings, priority ordering for available displays, and a conservative global shortcut to connect the first available display without replacing a known connection.
- Use the same native AppKit reorder behavior as the Installed page: available displays get a real drag preview and insertion gap; connected, unknown, and offline rows cannot be dragged.
- Localize all Sidecar copy for every repository-supported language and include portable Sidecar priority, connection, and shortcut settings in preference export/import.

## Notes

- SidecarCore is loaded dynamically at runtime; no private framework is statically linked.
- Wired-only mode remains a per-display connection policy and never falls back to Wi-Fi.
- Automatic switching is intentionally disabled when Sidecar connection state is unknown or ambiguous.
- Preference backups continue to exclude permissions, caches, credentials, and other non-portable data.

## Validation

- `xcodebuild -project MacTools.xcodeproj -scheme MacTools -configuration Debug -derivedDataPath build/DerivedData test -only-testing:MacToolsTests/SidecarPluginTests -quiet`
- `make generate`
- `make build-plugin PLUGIN=sidecar`
- `jq empty Plugins/Sidecar/Resources/Localizable.xcstrings`
- `git diff --check`


ggbond268:
I re-checked the two medium-severity concerns against the released v1.1.0 host and the Carbon runtime. Both are reproducible and should be addressed before merging.

### 1. The declared minimum host version does not cover the new PluginKit dependencies

The produced Sidecar bundle has strong undefined references to `PluginPortablePreferencesProviding` and `PluginSettingsReorderTableView`, both introduced by this PR. The released MacTools v1.1.0 host already uses PluginKit 3, but does not export either symbol. With `pluginKitVersion: 3` and `minHostVersion: 1.0.2`, v1.1.0 passes the manifest compatibility checks and then fails at `Bundle.load()`.

Please set `minHostVersion` to the first app release that contains these symbols, or bump the PluginKit compatibility version and publish Sidecar on that ABI line.

### 2. Sidecar shortcut registration bypasses host conflict validation

Sidecar registers Carbon hotkeys through its own `SidecarShortcutManager`, while the settings view only checks duplicates among Sidecar bindings. It cannot detect conflicts with host-level or other plugin shortcuts. I verified the platform behavior directly: registering the same combination twice returns `eventHotKeyExistsErr` (`-9878`) for the second registration. The current code only logs that failure after the UI has already persisted and accepted the binding, so depending on registration order either Sidecar or an existing MacTools shortcut silently stops working.

Please expose stable dynamic shortcut definitions through `PluginHost`, or add a host API for plugin-owned dynamic shortcuts, so registration and conflict validation remain centralized.

PR #191, opened 2026-07-15
Fix adaptive secondary panel placement

xcv58:
## Summary

- Make the shared secondary-panel presentation adapt to the available screen space.
- Keep long secondary lists usable by bounding their height and scrolling their content.
- Add placement coverage and a release-note entry.

## Root cause

The fixed-width secondary NSPanel was always positioned to the right of its source row without checking the display visible frame. When the menu-bar popover was near the right screen edge, the panel could render offscreen.

## Behavior

- Open on the right when it fits.
- Open on the left when the right side does not fit.
- Use an in-panel drill-in view with a Back control when neither side fits.

### Normal placement (right)

codex-clipboard-1a20f05c-bb59-458b-9048-483d51d39317

### Near-right-edge fallback (left)

codex-clipboard-6c1210b3-a26f-416b-9ac3-b0400145f0fb

### Wide-screen edge fallback (left)

codex-clipboard-bde2699f-20ed-438e-82cb-77f16015686a

## Validation

- `xcodebuild -project MacTools.xcodeproj -scheme MacTools -configuration Debug -derivedDataPath build/DerivedDataSecondaryPanel -only-testing:MacToolsTests/MenuBarPanelLayoutTests test -quiet`
- `make build`
- `make run`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 
 | Feature | Description |
 | ------- | ----------- |
-| Display Resolution | View connected displays and switch each display to an available resolution. |
+| Display Resolution | View connected displays and switch each display to an available resolution, with an adaptive list that remains usable near screen edges. |
 | Display Brightness | Quickly adjust built-in and DDC/CI external display brightness, with shortcuts that can follow the mouse or control all displays together, plus Gamma/Shade fallbacks. |
 | True Tone | Automatically adapt display colors to ambient light on MacBooks and compatible displays. |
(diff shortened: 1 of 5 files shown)


ggbond268:
The inline secondary-panel fallback introduced a Back button with hard-coded Chinese accessibility and help text. This has been fixed by routing the button through AppL10n and adding translations for all supported languages, so it now follows the app language.

PR #205, opened 2026-07-17
fix(mouse-enhancer): middle-click stops working after app restart

### LINKED ISSUE #202: [bug] 模拟鼠标中键需要重启开关才生效
重启 App 后，功能就失效了，需要重启开关，模拟鼠标中间功能才可以使用

Image

ayangweb:
fixed #202

After each app or system restart, the simulate-middle-click feature is silently broken until the user toggles its switch off and back on.

Two defects in `MouseEnhancerMiddleClickSession`:

### 1. CGEvent tap not re-enabled when macOS auto-disables it

macOS can send `tapDisabledByTimeout` / `tapDisabledByUserInput` to a non-listen `cghidEventTap` during busy startup or after sleep. The tap callback did not handle these events, so the tap stayed disabled permanently. `MouseEnhancerSession` (scroll reversing) already has this recovery; add the same to the middle-click session:

```swift
if type == .tapDisabledByTimeout || type == .tapDisabledByUserInput {
    if let tap = session.eventTap, CFMachPortIsValid(tap) {
        CGEvent.tapEnable(tap: tap, enable: true)
    }
    return passthrough
}
```

### 2. MTDeviceCreateList returns empty at login-item startup

When the app launches as a login item at boot, `MultitouchSupport.framework` may not have finished enumerating devices yet, so `_mtDeviceCreateList()` returns an empty array and no touch listeners are registered. The existing IOKit `kIOFirstMatchNotification` recovery does not help here—it drains currently-present devices on registration (arming future arrivals) but never fires a callback for devices already in the IOKit registry. Result: no touch data, `threeDown` never set, middle-click silently dead.

Schedule a one-shot 3-second rebuild at the end of `start()` when no devices are found, mirroring the existing wake-restart recovery:

```swift
if devices.isEmpty {
    scheduleRestart(after: 3, reason: "noDevicesAtStart")
}
```

### Tests

Added `testMiddleClickStartsOnActivateWhenSavedAsEnabledAndAccessibilityGranted` and `testMiddleClickDoesNotStartOnActivateWhenAccessibilityDenied` to cover the `activate()` path (app-restart scenario) with pre-persisted settings, which had no test coverage.


ayangweb:
@ggbond268 copilot 修复的，需要你来测试

PR #263, opened 2026-08-10
feat: add Dock Lock plugin

StephaneRavet:
## Summary
- add a Dock Lock plugin that prevents accidental Dock movement between displays
- require Accessibility permission and only clamp the pointer when the Dock is confirmed at the bottom
- add lifecycle, geometry, permission, and orientation tests

## Verification
- `git diff --check`
- plugin config generation
- JSON validation
- `swiftc -parse`

Xcode/xcodegen are unavailable in this environment, so the XCTest target was not run.


ggbond268:
Thanks for the contribution. I found three blocking issues:

1. The event tap cannot start for normal users
The code uses .cghidEventTap, which Apple documents as root-only. MacTools runs as a normal user, so CGEvent.tapCreate will return nil. Please use a supported user-session event tap and test the real monitor without root privileges.

2. The boundary logic blocks valid pointer movement
The implementation clamps the entire bottom edge of every display. This can prevent moving to a display positioned below another display and may also prevent an auto-hidden Dock from appearing. It should only clamp exposed edges that can trigger unwanted Dock movement.

3. The new test suite fails
testMissingPermissionDoesNotStartMonitor() creates a disabled context with makeContext(), so no permission error is expected. It should use makeContext(isEnabled: true). I ran the test class locally: 7 tests passed and this test failed.

PR #272, opened 2026-08-12
fix(plugins): validate manifest paths before sync mutations

### LINKED ISSUE #262: Validate plugin manifest paths before debug plugin sync mutations
## Summary

`scripts/plugins/sync-debug-plugins.sh` trusts `plugin.json.id` and `bundleRelativePath` after checking only that they are nonempty. Both values are used to build paths for destructive or mutating operations.

This was found while reviewing PR #256, but the affected manifest-handling lines predate that PR, so it should be tracked independently.

## Current behavior

The discovery code reads:

- `plugin_id = data.get("id") or ""`
- `bundle_relative_path = data.get("bundleRelativePath") or ""`

It then passes those values to shell code that constructs paths such as:

- `$PACKAGES_DIR/$plugin_id.mactoolsplugin`
- `$INSTALL_DIR/$plugin_id.mactoolsplugin`
- `$package_path/$bundle_relative_path`

Those paths are subsequently consumed by `rm -rf`, `mkdir -p`, `ditto`, and `mv`.

Because the script accepts an arbitrary `--source-dir`, a malformed local manifest can use `../` components or an absolute/otherwise non-normalized bundle path to escape the package or install roots.

## Expected behavior

Before the first filesystem mutation:

1. Validate plugin IDs with the same canonical grammar used by `PluginPackageManifestLoader`.
2. Require `bundleRelativePath` to be a normalized relative path with no absolute, empty, `.`, or `..` components.
3. Resolve package, staging, install, and bundle destinations and verify they remain descendants of their intended roots.
4. Fail without changing any path when validation fails.

## Regression coverage

Add malicious manifest fixtures for both `id` and `bundleRelativePath`. Place sentinel files immediately outside the package/install roots, run the sync command, and assert:

- the command fails;
- no sentinel is removed or modified;
- no directory or bundle is created outside the intended roots.

bunnysayzz:
## Problem

`scripts/plugins/sync-debug-plugins.sh` trusts `plugin.json` `id` and `bundleRelativePath` after checking only that they are non-empty, then uses them raw in `rm -rf`, `mkdir -p`, `ditto`, and `mv`. Because the script accepts an arbitrary `--source-dir`, a malformed local manifest can use `../` components or an absolute bundle path to escape the package and install roots (dev-time destructive operations).

## Fix

- **Validate during discovery, before any filesystem mutation.** Discovery (and its manifest validation) now runs before `OUTPUT_DIR`/`Packages`/state/install directories are created, so an invalid manifest aborts the sync without changing anything on disk.
- **Plugin IDs must match the `PluginPackageManifestLoader` grammar**: `[A-Za-z0-9][A-Za-z0-9._-]{1,126}[A-Za-z0-9]`, with the reserved `marketplace` value rejected.
- **`bundleRelativePath` must be a normalized relative path**: no absolute paths, and no empty, `.`, or `..` components.
- **Defensive destination containment**: package, staging, and install destinations are asserted to stay within their intended roots before any `rm`/`mkdir`/`ditto`/`mv` runs.

## Verification

- 7 new malicious-manifest regression tests (traversal `id`, reserved `marketplace` id, non-grammar id, `..`/absolute/`.`/empty-component bundle paths) with sentinel files placed directly outside the package and install roots: each run fails, reports the offending field, and leaves the sentinels untouched with no directory created outside the roots.
- Positive control: a valid manifest still syncs and installs end to end.
- Full script suite (`python3 -m unittest discover -s scripts/tests`): 18/18 pass.

Fixes #262

### DIFF
--- a/scripts/plugins/sync-debug-plugins.sh
+++ b/scripts/plugins/sync-debug-plugins.sh
@@ -90,17 +90,13 @@ if [[ -z "$SOURCE_DIR" || ! -d "$SOURCE_DIR" ]]; then
     exit 1
 fi
 
-PRODUCTS_DIR="$(cd "$PRODUCTS_DIR" 2>/dev/null && pwd || true)"
+PRODUCTS_DIR="$(cd "$PRODUCTS_DIR" 2>/dev/null && pwd -P || true)"
 if [[ -z "$PRODUCTS_DIR" || ! -d "$PRODUCTS_DIR" ]]; then
     echo "Debug build products directory not found. Run 'make build' first: $PRODUCTS_DIR" >&2
     exit 1
 fi
 
 REPO_ROOT="$(cd "$(dirname "$0")/../.." && pwd)"
-OUTPUT_DIR="$(mkdir -p "$OUTPUT_DIR" && cd "$OUTPUT_DIR" && pwd)"
-PACKAGES_DIR="$OUTPUT_DIR/Packages"
-CATALOG_PATH="$OUTPUT_DIR/catalog.dev.json"
-STATE_DIR="$OUTPUT_DIR/.sync-state"
 MANIFEST_COPY_SCRIPT="$REPO_ROOT/scripts/plugins/copy-plugin-manifest.py"
 APP_VERSION_CONFIG="$REPO_ROOT/Configs/AppVersion.xcconfig"
 DEVELOPMENT_HOST_VERSION="$(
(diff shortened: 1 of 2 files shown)


xcv58:
Thanks—manifest validation and mutation ordering look good, and the submitted tests pass locally. There is still a blocking containment gap: `assert_path_within_root` compares path strings without resolving them. A pre-existing `Output/Packages` symlink can redirect `package_path` outside `OUTPUT_DIR`; I reproduced the script succeeding and deleting a sentinel in the external target.

Please canonicalize the intended roots and resolve/check the package, bundle, staging, and install destinations before mutation, including protection against symlinked path components. Please also add a symlink-escape regression test and place the malicious-manifest sentinels at the actual paths the unsafe values would target.

bunnysayzz:
Thanks for the sharp catch, and for actually reproducing the symlink redirect before flagging it. You were right: the old `assert_path_within_root` compared path strings, so a pre-existing `Output/Packages` symlink pointing outside `OUTPUT_DIR` sailed through the check and `rm -rf` happily deleted the sentinel in the external target.

The fix is pushed to the same branch (commit `de2fd61`):

1. **Resolve before comparing.** `assert_path_within_root` now canonicalizes both the root and the destination via a `resolve_path` helper that follows symlinks in every existing component (stopping at symlinks so they get resolved or refused, never silently appended). Roots are canonicalized up front with `pwd -P`.
2. **Check against the right root.** Package, state, and catalog destinations are now validated against the canonical `OUTPUT_DIR` rather than `PACKAGES_DIR`/`STATE_DIR`, so a symlinked `Output/Packages` or `Output/.sync-state` cannot redirect writes outside the output root. Bundle path is checked against `PRODUCTS_DIR`, and bundle/staging/install destinations are checked before any `rm -rf`/`mkdir`/`ditto`/`mv` runs.
3. **Regression tests.** Three new symlink-escape tests plant pre-existing symlinks at `Output/Packages`, inside the package path (bundle subdir), and at the installed destination, each pointing at an external dir with a sentinel; the sync must refuse and leave the sentinel untouched. The malicious-manifest tests now also place their sentinels at the exact paths the unsafe values would target (e.g. `Output/evil.mactoolsplugin` for id `../evil`) and snapshot the whole temp tree before/after to prove zero mutation.

Full suite: `python3 -m unittest discover -s scripts/tests` passes 21/21, and a manual run confirms a valid manifest still syncs and installs end to end, including on repeat runs.

Happy to adjust if anything else jumps out on review.

xcv58:
Thanks—the canonical containment checks now block the original symlink escape, and all 21 submitted tests pass locally. One ordering gap remains: destination validation still happens after mutations begin.

The script creates `OUTPUT_DIR`, `Packages`, `.sync-state`, and `INSTALL_DIR` before entering the containment checks. It also checks `install_path` only after rebuilding the package and writing its state. I reproduced a rejected `Output/Packages` symlink leaving `.sync-state` and the install directory behind, and a rejected install-destination symlink leaving a fully built output package and fingerprint behind.

Please add a preflight phase that resolves and validates every package, bundle, state, catalog, staging, and install destination for all records before creating directories or changing files. The symlink regression tests should snapshot the whole temporary tree before and after rejection, as the malicious-manifest tests now do.

bunnysayzz:
Good catch, and thanks for reproducing the leftover-directory case before flagging it. You were right: containment checks were still running after mutations had begun, since the script created `OUTPUT_DIR`, `Packages`, `.sync-state`, and the install store before the checks, and `install_path` was only validated after the package was rebuilt.

The script now has a preflight phase that runs before anything touches the disk:

- Roots are canonicalized with `resolve_path` without creating anything (no `mkdir -p`).
- Every destination for every record is resolved and validated up front: package, bundle, bundle destination, state, staging, install, and catalog, all checked against their canonical roots (`OUTPUT_DIR`/`PRODUCTS_DIR`/`INSTALL_DIR`).
- Only after every check passes do the sync directories get created and mutations start.

So a rejected `Output/Packages` symlink now aborts with `.sync-state` and the install store never created, and a rejected install-destination symlink aborts before the package is rebuilt or its fingerprint written. The three symlink regression tests now snapshot the whole temporary tree before and after the rejected run and assert it is byte-for-byte identical, same as the malicious-manifest tests. All 21 tests pass locally, and a valid sync still works end to end including the idempotent second run.

xcv58 (approved):
Thanks for addressing both rounds of feedback. I rechecked the latest head and confirmed:

 - manifest validation runs before filesystem mutation;
 - package, bundle, state, catalog, staging, and install destinations are preflighted for resolved-path containment;
 - symlink escapes are rejected without leaving partial output;
 - a later unsafe plugin record cannot leave changes from an earlier valid record;
 - all 21 submitted tests pass locally.

No further blocking issues from my review. Approved, pending the GitHub Actions build passing.

PR #314, opened 2026-08-20
feat: add automatic preferences backups

### LINKED ISSUE #259: feat: add automatic settings backup and retention
## Background

MacTools currently supports manual preference export and import, but users have no local history to recover from an unintended change to plugin configuration, shortcuts, or display order. Automatic backups would make such changes recoverable; they do not replace manual exports for migration or sharing.

## Proposal

Create a local automatic-backup history using the existing `PreferencesBackup` payload, so automatic backups and manual imports cover the same portable preferences.

- Enable “Automatically back up settings” by default, while allowing users to turn it off.
- Create a snapshot 60 seconds after a meaningful persisted-preference change, and flush pending work before app termination.
- Do not write a new file when the meaningful content is unchanged; ignore timestamp fields such as `exportedAt` when comparing snapshots.
- Write atomically to `~/Library/Application Support/MacTools/Automatic Backups/`.
- Use GFS retention: keep all backups from the last hour; one per hour for the next 24 hours; one per day for the next 30 days; then one per week. Cap history at 100 snapshots and 25 MiB total.
- Write an immediate safety snapshot before an import overwrites the current settings.
- Ensure automatic backup files meet the same size and format constraints as importable files, so every stored snapshot is restorable.

## Settings UI

Extend the existing Preferences Backup section with:

- An automatic-backup switch, enabled by default
- “Back Up Now”
- “Open Backup Folder”
- A concise explanation that backups are stored locally and exclude permissions, caches, credentials, and other private data

The initial restore flow can reuse the existing import UI. A dedicated backup browser, diff view, and one-click restore can be separate follow-up work.

## Implementation notes

- Do not schedule backups solely from every `plugin.onStateChange` call, as that callback may also represent live-status refreshes. Introduce a central coordinator notified by host preference mutations, application-preference saves, and an explicit “persistent preferences changed” signal from plugins.
- Capture the snapshot on the main actor if required, but encode, write, and rotate it off the main thread so settings interaction remains responsive.
- Add coverage for deduplication, debounce behavior, termination flushing, pre-import snapshots, GFS retention, count/total-size caps, and write failures.

xcv58:
## Summary

- add enabled-by-default automatic local preferences backups with semantic deduplication, atomic writes, and GFS retention
- retain up to 100 snapshots or 128 MiB, with a 16 MiB limit for individual backup files
- trigger automatic backups only from explicit successful-persistence callbacks, including buffered plugin initialization migrations, and assemble the full snapshot after the 60-second debounce fires
- keep automatic, manual, import-safety, and termination writes revision-ordered; failed attempts preserve and re-arm dirty work instead of dropping the pending backup
- suppress semantic no-op signals and keep rejected persistence from scheduling backups across all six portable-preferences plugins
- add pre-import safety snapshots, termination flushing, localized privacy copy, and readable local date-time filenames while continuing to manage legacy snapshots
- separate automatic backup, manual backup, and preference transfer controls
- show relative backup freshness on the first line and the exact timestamp, snapshot count, and total size on the second; manual backups report whether a new snapshot was written or nothing changed

## Verification

- AutomaticPreferencesBackupTests: 32 tests passed
- PreferencesBackupChangeReporterTests: 7 tests passed
- full macOS XCTest suite
- 86 script unit tests
- PluginKit v5 binary compatibility check
- project generation and localization JSON validation
- changelog validation and git diff check
- fresh independent two-stage review at the verified 1,958-path fingerprint; final disposition: CLEAN
- signed local build installed and launched from `~/Applications/MacTools Dev.app`
- live English and Simplified Chinese UI verification, including relative/exact timestamps and the no-change indicator

### Preferences Backup settings (English)

Preferences Backup settings showing relative and exact backup timestamps plus the manual no-change result in English

### Automatic backup snapshots

Automatic Backups folder showing a readable local date-time filename

### Localized settings (Simplified Chinese)

Preferences Backup settings showing relative and exact backup timestamps plus the manual no-change result in Simplified Chinese

Closes #259


ggbond268:
The current implementation observes the entire UserDefaults domain, so runtime-only writes—such as Activity Stats, IP Overview cache, and workflow run history—can trigger unnecessary full backup checks.

This may cause avoidable CPU usage, disk reads, and UI pauses.

Expected behavior: Changes to data included in PreferencesBackup should schedule one debounced backup. Writes to runtime-only data, caches, or history should neither trigger nor postpone a backup. Multiple relevant changes should be coalesced, and termination should flush only pending relevant changes.

xcv58:
Good catch — this was true of the earlier implementation. Fixed in `12945e2d`.

The backup path no longer observes the entire `UserDefaults` domain. It now uses explicit callbacks emitted only after backup-relevant preferences are successfully persisted. Runtime-only data such as activity stats, IP cache data, and workflow run history does not trigger or postpone backups.

Relevant changes are still coalesced by the 60-second debounce, and the full snapshot is assembled only when that debounce fires. Termination flushes only when a relevant revision remains pending.

I also added coverage confirming:

- runtime history does not report a preference change;
- irrelevant defaults writes neither schedule nor postpone backups;
- repeated relevant changes are debounced;
- no-op or rejected writes do not schedule backups;
- termination without a pending relevant change does not create a snapshot.

The full local and GitHub Actions test suites are passing. Thanks for flagging this.

PR #318, opened 2026-08-20
Add Dock Click Hide plugin

StephaneRavet:
## Summary

- Add the Dock Click Hide plugin.
- Hide the active app after a plain click on its active Dock item.
- Provide persisted enablement, permission guidance, localized copy, and focused tests.

## Validation

- `make build-plugin PLUGIN=DockClickMinimize`
- `xcodebuild -project MacTools.xcodeproj -scheme MacTools -configuration Debug -derivedDataPath build/DerivedData -only-testing:MacToolsTests/DockClickMinimizePluginTests test -quiet`

Manual macOS Dock interaction checks remain to be completed.


ggbond268:
Thanks for the implementation. The overall design looks reasonable, A few items should be addressed:

1. A leftMouseDown is currently treated as a completed click, so dragging or long-pressing a Dock icon may also hide the application. Please confirm the action on mouseUp and cancel it after dragging or holding.
2. Non-Dock clicks unnecessarily invalidate and refresh the cached Dock PID. The cache should only be refreshed when the Dock process restarts.
3. Check that the clicked bundle matches the frontmost application before calling hasVisibleWindow, avoiding unnecessary Accessibility queries.
4. Accessibility calls currently run synchronously in the main event callback. Consider moving them to a serial background queue to avoid potential UI pauses when an application is unresponsive.

The performance impact should be small during normal use, but items 2 and 3 are straightforward optimizations worth including.

PR #335, opened 2026-08-23
feat: add secure Clipboard workspace with history and snippets

### LINKED ISSUE #306: feat(plugin): add privacy-first text clipboard history and pinned snippets
## Summary

Add an official `clipboard-history` plugin that stores a searchable, local history of copied text and supports pinned snippets.

The first version should be privacy-first and intentionally limited to text. Images, files, cloud synchronization, OCR, and cross-device sharing are out of scope.

Related: #95

## Motivation

MacTools currently provides Clipboard Clear but does not help users recover previously copied text or maintain reusable snippets.

Clipboard history is most valuable when it is:

- Easy to invoke from anywhere
- Clearly local
- Predictable about retention
- Easy to pause or erase
- Integrated with MacTools actions and shortcuts
- Explicit about what it observes and stores

## Product principles

1. **Local by default**
   - Clipboard content must not be uploaded.
   - No account or network service is required.

2. **Collect the minimum**
   - Capture text only in the MVP.
   - Ignore unsupported pasteboard types.

3. **User-controlled retention**
   - Users choose maximum item count and expiration.
   - Clearing history is immediate and verifiable.

4. **No hidden global search exposure**
   - Transient clipboard entries must not be added to the static Unified Search index.
   - Clipboard search happens inside the plugin's dedicated panel.

5. **Clear pause state**
   - The UI must visibly indicate when history collection is paused.

### History collection

Observe changes to the general pasteboard while the plugin is active.

Capture plain text when:

- The value is non-empty.
- The value is below the configured size limit.
- The same normalized value is not already the newest item.
- Collection is not paused.
- The likely source application is not excluded.

Store:

- Stable item ID
- Captured text
- Capture date
- Best-effort source application bundle identifier and name
- Pin state
- Last-used date

Source-application attribution is best effort. It should use the shared system-context service when available and must not be represented as guaranteed provenance.

### History browser

Provide a searchable floating panel or workspace with:

- Recent items
- Pinned snippets
- Search
- Copy
- Delete
- Pin/unpin
- Clear unpinned history
- Clear all history
- Pause/resume collection

Keyboard navigation should be supported.

Selecting an item should copy it back to the pasteboard.

Automatically synthesizing Command-V is not required for the MVP.

### Retention settings

Support:

- Maximum item count
- Expiration period
- Maximum text size
- Excluded applications
- Capture pause
- Clear history
- Clear pinned snippets separately

Suggested default behavior:

- Modest bounded history size
- Expiration enabled
- Oversized values ignored
- Pinned snippets exempt from ordinary expiration

Final defaults should be reviewed during implementation.

## Storage and privacy

Use a dedicated plugin store rather than `UserDefaults` for clipboard payloads.

Requirements:

- Encrypt clipboard payloads at rest using a key stored in Keychain.
- Do not write clipboard text to logs, crash breadcrumbs, analytics, or action-history details.
- Store only metadata needed by the product.
- Delete expired items from persistent storage rather than merely hiding them.
- Ensure uninstall removes the plugin database and associated encryption key.
- Redact content from diagnostics and support exports.

If encrypted storage cannot be completed in the initial implementation, the feature should not silently fall back to plaintext. The product must either block release or explicitly redesign the storage contract.

## Application exclusions

Allow users to exclude applications by bundle identifier.

Collection filtering should happen before clipboard content is persisted.

Because the pasteboard does not provide guaranteed source-application identity, the UI and documentation must state that exclusion uses the frontmost application observed at capture time.

The implementation should never claim that exclusions provide the same guarantees as a secure-input field or password manager.

## Proposed canonical actions

| Action ID | Behavior |
| --- | --- |
| `open-history` | Open the searchable clipboard-history panel |
| `pause-collection` | Pause future history capture |
| `resume-collection` | Resume capture |
| `toggle-collection` | Toggle capture state |
| `clear-unpinned-history` | Clear non-pinned history with confirmation |
| `clear-all-history` | Clear all history and snippets with stronger confirmation |

Do not create one canonical action or Unified Search result for every transient history item.

Pinned-snippet actions can be considered later after privacy, portability, and catalog-size behavior are defined.

## Proposed architecture

- `ClipboardChangeMonitoring`
  - Detects pasteboard ownership/change transitions.
- `ClipboardCaptureCoordinator`
  - Applies pause, exclusion, type, deduplication, and size policies.
- `ClipboardHistoryStore`
  - Owns encrypted persistence, retention, and deletion.
- `ClipboardHistoryQuerying`
  - Provides bounded search and pagination.
- `ClipboardSourceContextProviding`
  - Supplies best-effort frontmost-application context.
- `ClipboardHistoryPanelController`
  - Owns presentation and keyboard interaction.
- `ClipboardHistoryPlugin`
  - Publishes settings, state, permissions, and canonical actions.

Keep pasteboard monitoring separate from persistence so both can be tested independently.

## Implementation plan

- [ ] Define the history and retention data model.
- [ ] Implement an encrypted local store.
- [ ] Implement text-only pasteboard monitoring.
- [ ] Add deduplication and size limits.
- [ ] Integrate best-effort source-application context.
- [ ] Add application exclusions.
- [ ] Add retention cleanup.
- [ ] Build the searchable history panel.
- [ ] Add pin/unpin and snippet storage.
- [ ] Add pause/resume and clear controls.
- [ ] Publish canonical actions.
- [ ] Add uninstall cleanup.
- [ ] Add privacy documentation and Marketplace metadata.
- [ ] Add unit, integration, and manual tests.

## Test coverage

Include tests for:

- Empty clipboard
- Unsupported types
- Duplicate consecutive values
- Oversized text
- Rapid clipboard changes
- Clipboard ownership changing during capture
- Pause/resume
- Excluded applications
- Missing source-application context
- Retention expiration
- Maximum-count pruning
- Pinned-item retention
- Encryption-key loss
- Store corruption
- Uninstall cleanup
- No clipboard content appearing in logs

## Acceptance criteria

- [ ] The plugin captures supported text while enabled.
- [ ] Users can search, copy, delete, pin, and unpin entries.
- [ ] Collection can be paused and its state is clearly visible.
- [ ] Transient entries are not added to global Unified Search.
- [ ] History is bounded by count, age, and item size.
- [ ] Clipboard payloads are encrypted at rest.
- [ ] Clearing history removes persistent payloads.
- [ ] Excluded applications are filtered before persistence when source context is available.
- [ ] Uninstall removes clipboard data and cryptographic material.
- [ ] No clipboard text is written to application logs.

## Non-goals

- Images
- Files
- Rich-text previews
- OCR
- Cloud sync
- Cross-device clipboard
- AI classification
- Automatic secret detection
- Guaranteed password-manager protection
- Automatically typing or pasting into applications
- Exposing every history item as a canonical action

xcv58:
## Summary

Adds a privacy-first, keyboard-driven Clipboard workspace for finding earlier copies, keeping reusable content, creating snippets, and pasting several items across apps.

Closes #306.

## What users get

- **Encrypted local History.** Search text, rich text, images, PDFs, file references, colors, and supported audio/video representations. Image OCR makes recognized text, links, and email addresses searchable. Collection is opt-in and files remain references rather than archived copies.
- **One library model.** History, Saved clips, and editable Snippets share one encrypted store. Saving a captured clip changes metadata on the same item instead of duplicating its payload or result count; there is no separate Favorites layer.
- **A fast keyboard workflow.** Search supports short fragments (`88` matches `MT88 tripod`), progressive filters, paged lazy results, ordered multi-selection, customizable window commands, and a searchable companion Actions panel.
- **Paste, export, and share.** Paste original or plain text, paste recognized image text, replay file references in Finder, combine selected text in selection order, export supported formats, or use native macOS sharing.
- **Reusable snippets.** Create a snippet from scratch or from a clip, add tags and an optional expansion keyword, and use paste-time date/time, clipboard, environment, and cursor variables. The variable picker shows the template and a live example before insertion.
- **Sequential Paste.** A dedicated Paste Next shortcut snapshots recent History or uses an immutable ordered selection. A draggable HUD shows progress and preview content. New external copies end an implicit queue; explicitly selected queues retain their snapshots until used or dismissed.

The Clipboard window also supports Window Layouts, native resizing and movement from its visible handle, compact progressive filters, grouped shortcut help, and an Actions panel that stays associated with the Clipboard window without obscuring its content.

## Privacy and safety

- Password-manager and custom app exclusions, **Private Copy**, and **Ignore Next Copy** keep selected content out of History.
- Keychain stores only the encryption key; clipboard payloads remain in the local encrypted database.
- Item count, storage, age, per-item, and expanded-snippet limits are explicit and configurable. History supports up to 10,000 retained items; Saved clips and snippets are not removed by History retention.
- Keyword expansion is optional, skips secure fields, validates editor ownership and selection, and cancels stale replacement work when the user keeps typing.
- Clear and delete operations are confirmed, and the plugin declares its private-data uninstall policy.

<details>
<summary><strong>Reliability and concurrency coverage</strong></summary>

- History capture and snippet expansion use independent lazy pasteboard readers, so blocked snippet work cannot stall capture.
- Item-level merges preserve concurrent saves, captures, OCR, usage updates, and deletes without restoring stale snapshots.
- Paste, export, and companion-panel actions retain and revalidate their original targets before side effects.
- Usage bookkeeping is deduplicated and stays off the interaction path.
- Queue creation, progress, cancellation, History clearing, and app relaunch are revision-guarded so cancelled or stale work cannot recreate a hidden queue.
- Ordinary metadata updates patch affected visible rows without rerunning unrelated search or retention work.
- Reopening unchanged data restores the bounded initial page through revision tokens instead of an O(n) main-thread comparison.
- Clipboard-window shortcut conflicts offer **Swap**, **Replace**, or **Cancel**, with rollback if persistence fails.

</details>

<details>
<summary><strong>Performance evidence</strong></summary>

Frozen before/after binaries were measured in alternating Release runs. Median synthetic model results:

| Scenario | Before | After |
| --- | ---: | ---: |
| Unchanged reopen, 50,000 items | 57.2 ms | 0.015 ms |
| Metadata update through result completion, 50,000 items | 86.0 ms | 10.8 ms |
| Model notifications for that metadata update | 8 | 1 |

A separate within-build image test measured **133 ms uncached -> 13 ms cached** for ten preview visits, including the first decode. These are model and decoder measurements, not rendered-frame or end-to-end app timings; they exclude fixture creation, real Keychain/database access, and disk payload reads.

[Full results and limitations](https://github.com/xcv58/MacTools/blob/codex/issue-306-clipboard-history/docs/performance/clipboard-2026-08-28.md) · [Reproducible probes](https://github.com/xcv58/MacTools/blob/codex/issue-306-clipboard-history/scripts/benchmarks/README.md)

</details>

## Validation

This PR remains a **draft**; no review is requested yet.

- `make ci` passes on the current head: the complete XCTest suite, **213 repository script tests**, changelog and localization validation, manifest-schema checks, and the PluginKit v5 compatibility client.
- Stress and performance coverage includes 10,000-/50,000-item search, mutation, cold-open, warm-reopen, revision invalidation, retention, preview-cache, and queue-race cases.
- Final-review regressions cover cancellation during queue creation, cancellation racing a queue control, clearing an implicit queue, inactive queue controls, the idle keyword-expansion fast path, localized shortcut descriptions, and settings accessibility.
- A clean Debug build was signature-verified, installed, and relaunched. The installed executable hash matches the exact current-head build.
- GitHub's required Build workflow is the live source of hosted-CI status for the pushed head.

## Current-head screenshots

Captured from the installed build after the final review fixes. The Clipboard examples are filtered to one non-sensitive repository URL.

### Clipboard settings

Clipboard settings on the current head

### Clipboard workspace

Clipboard workspace on the current head

### Searchable Actions panel

Clipboard Actions panel on the current head

<details>
<summary><strong>Earlier state coverage: empty, small, mixed, image, and advanced settings</strong></summary>

These August 27 captures use a reset library populated with non-sensitive sample content. They document states not recreated for the final smoke pass; surrounding labels have since been refined.

**Populated History and rich-text preview**

Populated clipboard history with rich-text preview

**Empty History**

Empty clipboard history

**One copied item**

Clipboard history with one text item

**Mixed small library and neutral-background color preview**

Small clipboard library with a neutral-background color preview

**Image preview**

Mixed clipboard history with an image preview

**Paste Queue, snippets, and sensitive-copy settings**

Clipboard workflow settings

**Expanded Paste Queue options**

Expanded queue settings

**Retention and local data controls**

Clipboard retention limits and local data controls

**Configurable Clipboard-window shortcuts**

Configurable Clipboard-window shortcuts

</details>


xcv58:
Thanks for the detailed review. I rechecked every finding against the current PR head and addressed the issues that should be fixed in this PR.

### Fixed

- **CI / Quick Look sendability:** cancellation now owns an immutable request handle instead of capturing a non-Sendable Quick Look request in a `@Sendable` closure.
- **Selected-row contrast:** shared palette rows now use AppKit's semantic selected-text color, including high-contrast appearances, instead of hard-coded white.
- **Return-key selector behavior:** `insertNewlineIgnoringFieldEditor:` now selects the configured alternate action directly, while ordinary `insertNewline:` still uses the event modifiers. A regression test covers the AppKit selector path.
- **Clipboard summary cost:** history count, saved count, payload bytes, and the oldest unprotected item are cached and updated when the collection changes instead of being rescanned from panel-state getters.
- **Capture retention cost:** captures use an O(1) safety check and skip the full retention pass when no item could be expired or evicted. The full pass remains the fallback whenever capacity, expiration, or an in-flight durable mutation makes it necessary.
- **Startup responsiveness:** initial retention pruning now runs on the persistence worker. A settings-revision check reapplies changed retention settings before the result is published.
- **Secure input:** snippet expansion now checks macOS Secure Event Input before Accessibility classification and clears its buffer without inspecting the focused field. This has a regression test.
- **PluginKit compatibility inventory:** `pluginPaletteSelectableRow` is now inventoried at its introduction version, and the inventory parser/test covers public `View` extension APIs.

### Kept intentionally

- **Shortcut assignment while temporarily unavailable:** `canAssign` remains based on whether the action is an interactive command, not its momentary availability. This lets users configure a shortcut even when the queue or current selection is not active; execution still reports the real unavailable state. The code now documents this distinction, and the existing regression test verifies it.
- **Helper-process reaping:** the helper uses `Foundation.Process`, which reaps the child when it terminates. I also ran an isolated 25-process forced-termination probe and observed no remaining processes or zombies, so adding another wait/reaping mechanism would duplicate Foundation's ownership.
- **Queue HUD fallback timer:** this is a main-run-loop timer, so it is suspended during system sleep and does not wake the Mac. It only resumes the lightweight fallback check after wake, within at most 0.5 seconds.

### Follow-up rather than a partial fix here

- **Whole-database rollback/replacement detection:** this is a valid hardening topic, but a manifest stored beside or inside the same database cannot detect replacement with an older, internally valid database-plus-manifest pair. A meaningful solution needs an external Keychain-anchored store identity or monotonic generation plus an explicit backup/restore policy. That should be designed separately so legitimate restores are not mistaken for tampering.
- **Splitting the large clipboard panel file:** worthwhile for maintenance, but it is a structural refactor with no user-visible correction and would add review risk to an already large feature PR.

Validation on the updated head includes the full local `make ci` suite after merging current `main`, the Clipboard History plugin build after the final capture-path cleanup, and the successful [current-head GitHub Actions Build](https://github.com/ggbond268/MacTools/actions/runs/33976820355).

PR #338, opened 2026-08-23
[Superseded][Prototype] Add signed local mactools CLI and capability API

xcv58:
> [!IMPORTANT]
> This broad prototype has been superseded by the narrower, review-ready Phase 0 transport PR #340. It is retained as design and implementation reference for later CLI phases and is not intended to be merged in its current form.

## Prototype status

This draft PR is a working prototype for RFC #309. It is intended to make the packaging, security, IPC, lifecycle, and command-surface decisions testable; it does not propose freezing every implemented command into CLI v1.

## What this prototype includes

- A separately downloadable, signed universal `mactools-cli` release archive whose installed command is `mactools`; the CLI executable is not embedded in `MacTools.app`.
- An opt-in, user-scoped LaunchAgent/XPC broker provided and registered by the app.
- A narrow broker responsible for peer authentication, protocol negotiation, host startup, forwarding, admission bounds, and cancellation.
- Host-owned action discovery and execution through the existing Action Registry and Action Executor.
- Exact role identity, Team ID, Apple anchor, and effective-user trust checks, with separate development service identities.
- Versioned request/response contracts, bounded cold start, structured JSON output, stable exit codes, cancellation, and upgrade/failure recovery.
- A broad prototype command surface covering action discovery/execution, typed and sensitive input transport, workflow conveniences, and read-only plugin diagnostics.
- Separate release packaging, notarization, checksums, artifact validation, and Homebrew-ready distribution support.

## Decisions reflected from the RFC discussion

- The CLI is optional and distributed separately while sharing the app release cadence.
- The app provides the broker; the CLI package contains only the command-line executable.
- The broker does not load plugins or execute actions.
- MacTools remains the authoritative source for installed plugins, availability, permissions, confirmation, concurrency, timeouts, and execution.
- Run Links remain supported and no network or public remote-control API is introduced.

## Areas to validate before defining v1

- Which of the broader prototype commands should remain in the first public release.
- Whether CLI eligibility should be fully independent from Run Link `externalInvocationPolicy` or retain it as an additional conservative gate.
- Installation, upgrade-ordering, background-item approval, and removal ergonomics.
- The appropriate initial policy for safe, interactive, confirmation-required, and parameterized actions.

## Verification

The branch includes unit, executable, protocol, packaging, release-layout, and end-to-end coverage. The latest response-contract changes were verified with:

```sh
xcodebuild -project MacTools.xcodeproj -scheme MacTools -configuration Debug \
  -derivedDataPath build/DerivedData test -quiet \
[...]

Issue #309, opened 2026-08-19
RFC: add a signed local CLI and capability API over canonical actions


xcv58:
## Summary

Introduce an official `mactools` command-line tool backed by a versioned, authenticated local IPC contract

The CLI should discover and invoke the same canonical actions used by Unified Search, global shortcuts, Automation, Action Grid, gestures, Run Links, and App Intents.

Canonical actions remain the domain model. Run Links remain the lightweight deep-link interface. The CLI becomes an additional programmatic surface with structured input, output, completion, and error behavior.

## Decision Requested

This RFC asks the project to decide:

- Whether MacTools should ship an official CLI
- Which local IPC architecture should connect the CLI to the running host
- How CLI action eligibility relates to existing external-invocation policy
- How the CLI locates or starts the MacTools host
- How the binary is distributed and placed on the user's shell path
- Which commands and guarantees belong in the first version

This RFC does not authorize a network server or remote-control API.

## Motivation

Run Links are appropriate for:

- Clickable integrations
- Apple Shortcuts
- Launchers
- Documentation
- Fire-and-forget shell commands
- Stored parameter presets

They are not sufficient for:

- Listing installed actions
- Describing action parameters
- Querying availability
- Passing typed parameters directly
- Receiving structured results
- Returning reliable shell exit codes
- Waiting for completion
- Cancelling cancellable work
- Machine-readable scripting
- Future local agent integrations

A CLI would make MacTools useful from:

- Terminal scripts
- Shell aliases
- Developer tooling
- Alfred, Raycast, Hammerspoon, and similar tools
- Test automation
- Local agents
- A future MCP adapter
- Support and diagnostic workflows

## Architectural Invariants

- Plugins continue to publish canonical actions through the existing action registry.
- The CLI does not load plugin bundles.
- The CLI does not implement plugin behavior.
- The CLI does not bypass availability, permission, confirmation, concurrency, or timeout policies.
- The host process remains authoritative for installed plugins and runtime state.
- The CLI communicates only through a local authenticated channel.
- No TCP port or local HTTP server is introduced.
- Run Links remain supported.
- Existing action references remain stable.
- Sensitive action values are never placed in URLs or ordinary command-line arguments.

## Proposed Command Surface

The initial command groups should be:

### General

- `mactools version`
- `mactools doctor`
- `mactools help`

### Actions

- `mactools actions list`
- `mactools actions describe <provider/action>`
- `mactools actions availability <provider/action>`
[...]

PR #340, opened 2026-08-24
[Prototype][Phase 0] Validate CLI transport and broker security

xcv58:
## Summary

This extracts the transport foundation from the broader CLI prototype into a narrow Phase 0 review surface.

- keeps `mactools` as a separately distributed executable rather than embedding it in `MacTools.app`
- adds an app-provided, user-opt-in LaunchAgent broker
- authenticates CLI, broker, host, and containing app by user, Apple signing anchor, Jenny Media team, and exact role identifiers
- negotiates an independent protocol range and supports only `help`, `version`, and `doctor`
- gives `doctor` one 10-second monotonic deadline, broker-owned cold launch, cancellation, and bounded admission
- applies capped exponential host reconnect backoff, caches only the validated local identity, and ignores stale callbacks from older broker connections
- reserves bounded cleanup time inside the existing `doctor` deadline, attempts explicit cancellation, and always invalidates terminal XPC state

## Explicit non-goals

No actions, workflows, plugins, PluginKit changes, parameter schema, installer, notarization, or public CLI release packaging are included.

## Verification

- full local `make ci` passes
- GitHub Build and test passes on macOS 26
- 23 focused reconnect, signing-boundary, deadline, cancellation, and lifecycle tests pass
- focused protocol, approval, admission, and executable tests pass
- clean Jenny Media-signed builds pass strict code-signature validation
- signed end-to-end testing passes on macOS 26.6.2 arm64 and macOS 27.0 arm64
- `version --json` succeeds with the host closed and does not launch it
- `doctor --json` cold-launches the exact containing app without stealing focus
- host and broker remain healthy with no observed crash report
- `MacTools.app` embeds the broker but not `mactools`

On macOS 26, the standalone CLI must be copied to a stable installed path. A byte-identical executable run directly inside the extracted Downloads test-kit directory could not establish the XPC broker connection. Distribution and installation UX remain later-phase work.

## Review and merge status

This prototype is ready for architecture and code review. Signed macOS 14 and macOS 15 runtime compatibility evidence remains a merge gate because the repository deployment target is macOS 14.

Related to #309. Narrow follow-up to the broad prototype in #338.


ggbond268:
A couple of non-blocking follow-ups:
1. If the CLI broker remains unavailable while the service is enabled, the host retries every second on the main actor, including code-signing validation. This could cause unnecessary CPU wakeups and occasional UI delays. Consider using capped exponential backoff, resetting it after a successful connection, and avoiding repeated signing checks on the main thread.

2. When a request reaches the overall timeout, cancellation reuses the already-expired deadline, so the explicit cancel request is skipped. Connection teardown should still clean up the current doctor request, so the practical impact is small. Consider using a separate short cleanup deadline for cancellation, followed by connection invalidation.

xcv58:
Addressed both non-blocking follow-ups in a7a3dbf2:

- Host reconnects now use capped exponential backoff (1s through 30s), reset only after an authenticated valid registration. The host caches its successfully validated local signing identity, while every broker connection is still independently verified. Connection generations prevent stale callbacks from resetting backoff or disrupting a newer connection.
- `doctor` now reserves 250 ms for cleanup inside the existing absolute 10-second deadline. Timeout/cancellation attempts explicit request cancellation when time remains, then always invalidates XPC state. Terminal failures during host preparation also invalidate the connection; successful preparation preserves it for the request.

Verification: 23 focused tests pass, full local `make ci` passes, and three independent two-stage review rounds ended clean after resolving the first round's lifecycle findings.

PR #342, opened 2026-08-26
Window Layouts: add modifier-key window dragging

xcv58:
## Summary

- add opt-in modifier-key window dragging to Window Layouts, targeting the topmost movable window beneath the pointer without clicking or changing focus
- support configurable exact modifier combinations, a 4-point activation threshold, host-window targeting, gesture conflict claims, and localized settings and status feedback
- serialize and coalesce pointer events with lifecycle generations so stale actions cannot survive stop or restart, and make Accessibility window scans cancellation-aware
- persist and reset the new settings, and document the feature in both READMEs and an unreleased plugin changelog fragment

## Validation

- complete Window Layouts test set passed through `MacToolsTests`
- focused regressions cover modifier release, conflict idempotence and resume, action coalescing, stale-action invalidation, Accessibility cancellation, no-window-under-pointer messaging, persistence, reset, and error clearing
- Window Layouts plugin-only Debug build passed
- `git diff --check`, localization JSON validation, plugin manifest JSON validation, and changelog validation passed
- built plugin package was installed into MacTools Dev and verified byte-for-byte against the package output; the app restarted successfully
- Xcode emitted only pre-existing Disk Clean actor-isolation warnings

## Manual testing remaining

- exercise modifier dragging across different apps, including apps with unusual Accessibility behavior
- test multiple displays and different display coordinate arrangements
- test rapid movement with high-polling-rate mice and modifier release or cancellation
- confirm empty-desktop error feedback clears after the next successful drag

## Manual verification

Tested locally: modifier-key window dragging works as expected.

Window Layouts modifier-drag settings


ggbond268:
Thanks for the implementation. I found two issues worth addressing:

1. Event-monitor startup failures are always reported as an Accessibility error, even though Accessibility has already been verified before start() is called. If event-tap or run-loop-source creation fails for another reason, the feature remains enabled and continues advertising its gesture claim while showing misleading permission guidance. Consider returning a Result from start(), using a distinct monitor-start error, and tracking whether the session is actually running.

2. While the feature is enabled, every global mouseMoved event is handled on the main run loop, even when the required modifiers are not pressed. With a high-polling-rate mouse, this may cause unnecessary CPU usage, energy impact, and UI-thread contention. Consider running the event tap on a dedicated thread, limiting movement processing to the armed/active state, and coalescing AX position updates to a bounded refresh rate.

Issue #312, opened 2026-08-19
build(metadata): generate catalogs and website plugin pages from plugin.json


xcv58:
## Summary
Use each official plugin’s existing `plugin.json` as the single checked-in source for:
* Package metadata
* Marketplace product metadata
* Capability metadata
* Discovery metadata
* Privacy and requirements disclosures
* Website plugin pages
* Website action pages
* Search metadata

Generate the signed Marketplace catalog and website data from that source. Do not require a separate `marketplace.json`. This issue builds on the capability-metadata schema defined by the catalog metadata issue and provides the generation pipeline needed by #308.

## Decision
Each official plugin has one source manifest: `Plugins/<Plugin>/plugin.json`.
That manifest may contain:
* Runtime and package fields
* Build fields
* Product-presentation fields
* Discovery fields
* Requirements
* Privacy declarations
* Static action descriptors
* Dynamic action templates
* Setup metadata
* Relationships

Generated outputs may be separate, but they must not become independently hand-maintained sources.

## Motivation
MacTools currently has several representations of plugin information:
* `plugin.json`
* Swift runtime metadata
* Generated Marketplace catalog
* Website TypeScript types
* Website category mappings
* Website presentation code
* Plugin documentation
* Runtime action definitions

Without a single source and validation pipeline, these can drift. **Possible failures include:**
* Website lists an action removed from runtime.
* Marketplace omits a permission documented elsewhere.
* Category labels differ between app and site.
* Privacy behavior is updated in one location only.
* A plugin release appears in the catalog but lacks a useful website page.
* An action rename leaves stale search metadata.
* A related plugin ID no longer exists.

## Source-of-Truth Principles
* `plugin.json` is the **only** required checked-in structured metadata file per plugin.
* Runtime Swift code remains authoritative for live action behavior.
* Build tools validate static metadata against runtime definitions.
* Generated files are not edited manually.
* Website builds are deterministic from repository content.
* Release catalog generation does not require fetching the production site.
* Presentation assets remain separate files referenced by the manifest.
* The signed catalog contains only the validated projection needed by clients.
* Build-only values do not need to appear in website output.
* Website-only presentation fields do not need to be consumed by the runtime host.

## Source-Manifest Model
Introduce a build-tool model representing the complete `plugin.json` schema. The model should contain the existing fields plus optional namespaced sections such as:
* `presentation`
* `discovery`
* `requirements`
* `privacy`
* `actions`
* `onboarding`
* `relationships`

[...]

PR #362, opened 2026-08-29
feat(site): generate plugin and action pages from manifests

xcv58:
## Summary

Generate website plugin and static-action pages from validated `Plugins/*/plugin.json` metadata and translations, replacing the production-catalog fetch with committed data and verified assets. Build and Pages workflows reject stale generated output.

- Preserve existing interactive settings previews and load shared language/theme controls once on every page.
- Keep setup layout styles on their outer wrappers so instructions display only the selected language.
- Regenerate metadata against current main, including Siri and the updated Xcode Cleanup requirements.

Open in MacTools links navigate to Marketplace details; they do not install plugins or execute actions. Signed release catalogs and runtime behavior are unchanged. Runtime/source action comparison and additional interactive preview fixtures remain outside this PR.

Refs #312.

## Validation

- Five generator tests and four website tests passed.
- Generated-data freshness check and website build passed (192 pages).
- Browser verification confirmed setup instructions display correctly in both Chinese and English, with their layout preserved.
- Changelog validation and `git diff --check` passed.

## Screenshots

Marketplace list:

MacTools marketplace

Window Layouts detail:

Window Layouts plugin details


ggbond268:
Thanks for working on this! I rechecked the latest commit and confirmed two issues to address before merging, plus one preview change to clarify:

1. **Language and theme switches do not work on the new detail pages.** I reproduced this on both plugin and action detail pages; the existing plugins page works correctly. The new pages render the navigation controls but do not load `site-controls.ts`. Please load the shared controls on both page types, ideally through the shared layout, and verify that both switches work.

2. **Existing website settings previews are replaced by a generic enable toggle.** For example, Fan Control no longer shows its preset and speed-slider previews. This only affects the website, not the app. I suggest keeping `settingsModels[plugin.id]` where available and using `fallbackSections` for plugins without a preview, until equivalent generated previews are ready. If this simplification is intentional, could you clarify the intended scope?

3. **Generated metadata is stale after merging the latest `main`.** The check passes on the PR branch, but a local merge with current `main` reproduces `stale generated output: .../plugins.json`. Since Pages runs this check before building, it would block the next deployment. Please sync with `main`, regenerate and commit the metadata, then rerun the freshness check and site build.

Items 1 and 3 need fixing before merging; item 2 is a scope decision rather than an app bug.

PR #364, opened 2026-08-30
fix(input-remapping): diagnose Custom Shortcuts recorder failures

### LINKED ISSUE #346: fix(input-remapping): surface recorder startup failures
## Summary

Custom Shortcuts can silently fail to enter recording mode when its global event tap cannot start. From the user’s perspective, clicking **Record input** or **Record shortcut** may appear to do nothing.

This issue is limited to failure reporting and recovery for the existing input/shortcut recorders. The single-key picker work in #343 is related input infrastructure but is not the behavior being requested here.

## Reported behavior

- Open **Custom Shortcuts: Keyboard, Trackpad, Mouse**.
- Add or edit a mapping.
- Click **Record input** or **Record shortcut**.
- In an affected permission/event-tap state, the UI remains unchanged and does not explain why recording did not start.

Source inspection identifies a matching silent failure path; the exact runtime permission configuration still needs to be reproduced during implementation.

## Source findings

- `InputRemappingButtonCaptureCoordinator.start(...)` and `startShortcut(...)` return `false` when `tap.start()` cannot create or activate the event tap.
- The corresponding settings actions discard that result with `_ = buttonCapture.start(...)` / `_ = buttonCapture.startShortcut(...)`.
- Consequently, an immediate startup failure does not become recorder state, an inline error, or permission guidance.
- Existing configuration errors mainly cover enabled runnable rules; they do not give the user feedback for this failed recording attempt.

Relevant files:

- `Plugins/InputRemapping/Sources/InputRemappingPlugin.swift`
- `Plugins/InputRemapping/Sources/InputRemappingEventTap.swift`
- `Plugins/InputRemapping/Tests/InputRemappingModelsTests.swift`

## Expected behavior

When recording cannot start, Custom Shortcuts should:

1. Show a clear, local error instead of silently returning to the idle UI.
2. Distinguish missing Input Monitoring or Accessibility permission from a generic event-tap failure where possible.
3. Offer the appropriate permission action or a retry action.
4. Preserve the existing mapping/draft without partially enabling it.
5. Return to a stable idle state after cancellation, Settings dismissal, app deactivation, or a failed retry.

## Suggested acceptance criteria

- A `tap.start() == false` path is visible and actionable in the rule editor.
- Permission-denied states point to the correct System Settings permission.
- A non-permission event-tap failure provides a retryable message.
- Starting another recorder or closing the settings page clears the failure/preparing/listening state safely.
- Focused tests cover startup failure, retry, cancellation, and draft preservation.
- Manual QA covers a real denied-then-granted macOS TCC flow, since mocked taps cannot validate system permission behavior.

## Original-author input

@StephaneRavet, since you authored the original Custom Shortcuts implementation and its guarded recording flow, could you share your opinion on the intended UX here—particularly whether recorder startup errors should remain inside each mapping card and whether there were constraints behind keeping the capture coordinator plugin-local?

StephaneRavet:
## Summary
- Surface typed recorder startup failures in the affected mapping and target.
- Show the relevant permission guidance or a retry action without changing the draft.
- Keep the capture coordinator plugin-local and cover immediate and delayed startup failures.

Closes #346

## Verification
- `make build-plugin PLUGIN=InputRemapping`
- Targeted `InputRemappingModelsTests`
- `make script-tests` (195 tests)
- `make ci` reaches the repository suite; unrelated existing failures remain in Action Grid, Plugin Manifest, Device Battery, and Disk Clean tests.


ggbond268:
Thanks for improving the recorder failure experience. The inline error and recovery actions are a good direction.

Besides resolving the merge conflicts, could you please address the following:

1. Restore the previous mapping when recording fails.
    For example, if a user changes an existing “Mouse Forward” action to “Shortcut” and recording cannot start, the rule may remain changed to an incomplete shortcut and become disabled. Please preserve the previous action and enabled state before recording, then restore them on both immediate and delayed startup failures while keeping the inline error visible.

2. Clear stale errors after the user changes the mapping.
    If the user selects another input or output after a recording failure, the old error may remain visible. Please clear the relevant error when the input/output changes or the rule is deleted.

Minor follow-ups:

* Avoid calling the capture cancellation path twice after a delayed startup failure.
* The Japanese and Korean translations currently read more like “audio recording”; wording closer to “input capture” or “input recording” would be clearer.

The overall approach looks good, and I did not find notable security or performance concerns.

StephaneRavet:
Conflicts have been resolved and the branch has been updated on top of the latest `main`. GitHub now reports this PR as mergeable. Targeted build and tests pass.

ggbond268:
Thanks for addressing the previous feedback. There is one remaining issue in the retry flow:
1. Change an existing action, such as “Mouse Forward,” to “Shortcut.”
2. Recording fails, correctly restoring the previous action.
3. Click “Try again” after resolving the startup issue.
Retry starts shortcut capture, but the recording controls are still hidden because the restored action is not a shortcut. Input can therefore be captured without a visible recording indicator or Cancel button.
Please show shortcutRecordingControl whenever this rule is preparing or recording a shortcut, regardless of its current action type. Keep the previous mapping available for rollback if recording fails or is cancelled.
Please also add a regression test covering failure → restore a non-shortcut action → retry, verifying that the recording state and Cancel button are visible.

StephaneRavet:
Addressed in 4c8d9e38.

- `shortcutRecordingControl` now stays visible whenever this rule is preparing or actively recording a shortcut, even after a failed attempt restored a non-shortcut action.
- The existing output snapshot is retained across retry, so failure or Cancel still restores the prior mapping.
- Added a regression test covering failure → non-shortcut restoration → retry, including both preparing and recording states and the visible Cancel path.

Verified with `make build-plugin PLUGIN=InputRemapping` and the targeted `InputRemappingModelsTests.testShortcutRetryAfterFailureShowsRecordingControlsForRestoredNonShortcutAction` XCTest. GitHub reports the PR as mergeable.

PR #374, opened 2026-08-31
feat(settings): add unified permission center and improve sidebar navigation

### LINKED ISSUE #372: feat(permissions): add a unified permission center backed by PermissionFlow
## Summary

Add a host-owned permission center and guided authorization coordinator for MacTools, using [PermissionFlow](https://github.com/jaywcjlove/PermissionFlow) as the System Settings guidance layer where it fits.

The goal is one coherent user journey across installed plugins without making a third-party package the source of truth for every permission type.

## Why this is needed

MacTools already has a useful permission abstraction:

- `PluginPermissionRequirement`
- `PluginPermissionState`
- host-rendered `PluginPermissionCard`
- plugin-owned `handlePermissionAction(id:)`

Today those cards are shown inside individual plugin settings pages and their status/request logic is mostly plugin-local. Current source contains 23 declared permission requirements across 16 plugin implementations, including 11 Accessibility requirements and 5 Input Monitoring requirements. A user with several such plugins sees the same host permission explained and opened repeatedly, with no single place to understand which installed features depend on it.

There are also permission-like flows outside the shared model, such as Disk Clean's Full Disk Access guidance.

## PermissionFlow feasibility

PermissionFlow is a reasonable dependency candidate:

- MIT licensed and actively maintained.
- Current tag at analysis time: `v2.11.2`.
- Supports macOS 13+; MacTools supports macOS 14+.
- Uses SwiftUI + AppKit and exposes a reusable `PermissionFlowController`.
- Provides guided panes for Accessibility, Full Disk Access, Input Monitoring, Screen Recording, Camera, Microphone, Calendar, Reminders, Bluetooth, Media & Apple Music, App Management, and Developer Tools.
- Provides `SystemSettingsKit` for typed deep links beyond the built-in guided panes.
- Allows status providers to be registered or replaced for known panes.
- Includes Simplified Chinese and English localization and a safe packaged-resource lookup API.

However, it is not a complete drop-in replacement for MacTools permission semantics:

- `PermissionFlowPane` is a closed set; arbitrary settings destinations can be opened by `SystemSettingsKit`, but they do not automatically gain the floating flow or status support.
- Automation permission is app-target-specific and normally appears only after the corresponding Apple Event request.
- System Audio Recording is distinct from Screen Recording and has different OS/version behavior.
- Full Disk Access cannot reliably report revocation until process restart; MacTools already has task-specific probing and restart guidance.
- Calendar permission should continue to use the plugin's EventKit request and status handling.
- The package's floating panel tracks the System Settings window through AX/Window Server behavior, so installed-app, multi-display, HiDPI, and permission-denied states require real QA.

## Architecture

Introduce a host-owned coordinator in App/Core. Do not expose PermissionFlow types through `MacToolsPluginKit` or require dynamic plugins to link the package.

Illustrative responsibilities:

```text
PluginPermissionRequirement / plugin-owned state
                    ↓
       MacTools PermissionCoordinator
          ↙ supported adapter  ↘ custom adapter
   PermissionFlow controller    plugin/host action
          ↓                         ↓
 guided System Settings      EventKit / Apple Events /
 flow and typed links        System Audio / task-specific UX
```

The coordinator should:

- Aggregate requirements from installed plugins.
- Deduplicate shared host permissions while listing the affected plugins/features.
- Keep plugin-provided descriptions and `PluginPermissionState` as the initial status authority.
- Route supported actions through one reusable PermissionFlow controller.
- Route unsupported/special cases through explicit MacTools adapters.
- Refresh states when MacTools becomes active after the user returns from System Settings.
- Support a sequential, user-driven review flow without prompting for everything at app launch.
- Let existing plugin settings cards deep-link into the same coordinator.

## Permission mapping

Initial mapping should be explicit and tested:

| MacTools requirement | Guidance path | Status/request authority |
| --- | --- | --- |
| Accessibility | PermissionFlow `.accessibility` | Existing plugin/host AX check initially |
| Input Monitoring | PermissionFlow `.inputMonitoring` | Existing plugin check initially |
| Screen Recording | PermissionFlow `.screenRecording` | Existing plugin preflight initially |
| Calendar Full Access | System prompt, then PermissionFlow/SystemSettingsKit calendar destination | Calendar plugin/EventKit |
| Automation | SystemSettingsKit Automation destination when repair is possible | Owning plugin and target app behavior |
| System Audio Recording | Dedicated Audio Capture destination/adapter | App Volume's existing macOS-specific state |
| Full Disk Access | PermissionFlow guidance may be used, but keep task-specific restart and coverage state | Disk Clean's existing probe/policy |

Do not continue representing System Audio Recording as generic `.screenRecording` merely to reuse button copy. Add a distinct host mapping or adapter without breaking PluginKit compatibility.

## User experience

Add a top-level **Permissions** destination in Settings:

- Show only requirements relevant to installed plugins and host features.
- Group duplicate host permissions into one card.
- List the plugins/features affected by each permission.
- Put missing or attention-needed permissions first.
- Explain whether the permission is required, optional, requested on first use, or requires an app restart.
- Offer one action at a time: request, open settings, recheck, or continue.
- Never imply that MacTools can grant a macOS permission itself.
- Never block unrelated plugins because one permission was denied.

After a Marketplace install or preferences restore, the completion UI may offer **Review required permissions** and open this destination. It must not automatically prompt for all permissions.

## Relationship to existing issues

- #307 is a permission-aware system-context broker and explicitly leaves a centralized permission coordinator as separate future work. Keep that issue focused on context acquisition; this issue may consume its structured availability later.
- #308 owns rich Marketplace details and continuous post-install setup. It should link to this permission center rather than implementing a second permission flow.
- #346 keeps Input Remapping recorder errors local to the affected mapping. Its permission-repair button can route through this coordinator without moving the error out of the card.
- #310 established catalog permission metadata for pre-install disclosure; runtime installed-plugin state remains authoritative here.

## Dependency and packaging requirements

- Pin a reviewed exact PermissionFlow version in `project.yml`, consistent with the repository's controlled package dependencies.
- Keep the dependency on the host app target; do not add it to every plugin bundle.
- Include the upstream MIT license/copyright notice in distributed third-party notices.
- Verify the PermissionFlow resource bundle is present and loadable in an installed signed Release app.
- Use `PermissionFlowResources` rather than assuming `Bundle.module` from host code.
- Start with `promptForAccessibilityTrust: false`; opening a non-Accessibility permission guide must not itself trigger Accessibility authorization.
- Evaluate only the optional status products MacTools actually needs; do not link the all-in-one extension by default without justification.

## Acceptance criteria

- [ ] Settings has one permission center derived from installed plugin requirements.
- [ ] Shared permissions are deduplicated and show every affected installed plugin/feature.
- [ ] Existing plugin settings cards route to the same coordinator.
- [ ] Accessibility, Input Monitoring, and Screen Recording use PermissionFlow guidance through a host adapter.
- [ ] Calendar, Automation, System Audio Recording, and Full Disk Access preserve their required specialized behavior.
- [ ] The app never prompts for every permission at launch or immediately after installing a plugin.
- [ ] Returning from System Settings refreshes visible status without logging or persisting sensitive data.
- [ ] Denying or skipping one permission does not prevent unrelated plugins from working.
- [ ] Dependency types do not cross the PluginKit ABI boundary, and existing dynamic plugins continue to load.
- [ ] The exact dependency version and required license notice are committed.
- [ ] An installed signed Release app can present localized PermissionFlow UI without a missing-resource crash.
- [ ] Manual QA covers macOS 14 and current macOS, denied/granted/revoked states, restart-required Full Disk Access, multiple displays, and HiDPI window placement.
- [ ] Tests cover requirement aggregation, deduplication, ordering, adapter mapping, unsupported/custom flows, app-activation refresh, and plugin removal/update.
- [ ] PluginKit-facing changes run `make script-tests`; the host integration passes targeted XCTest and a Release build.

## Non-goals

- Bypassing macOS TCC or granting permissions automatically
- Prompting for permissions unrelated to installed features
- Replacing task-specific availability/error UI
- Making PermissionFlow a public PluginKit dependency
- Folding the system-context broker from #307 into this work
- Moving pre-install permission disclosure out of #308

xcv58:
## Summary

- add a host-owned Permissions page that groups installed-plugin requirements, deduplicates shared permissions, and shows every affected feature
- request permissions only after an explicit user action; use PermissionFlow guidance for Accessibility, Input Monitoring, Screen Recording, Automation, and Full Disk Access while retaining the specialized Calendar, Finder extension, and System Audio flows
- refresh permission state when MacTools becomes active and expose the page through Settings navigation, search, keyboard shortcuts, and `mactools://app/settings/permissions`
- redesign Settings navigation with collapsible MacTools, Customize, and Plugin Settings sections; plugin sorting and filtering; selection history; and automatic selection reveal
- assign `⌘1…⌘9` to the current visible navigation targets, reveal their aligned hints only while Command is held, keep `⌘K` for global Settings search, and use `⌘⇧F` for plugin filtering
- declare Disk Cleanup's optional Full Disk Access requirement and include localized copy, documentation, changelog entries, tests, and third-party notices

### Unified permission overview

Permissions overview

### Dynamic, aligned navigation shortcuts

Number shortcuts are revealed while Command is held and follow the currently visible sidebar destinations.

Dynamic sidebar shortcuts

### Plugin settings filter and selection states

`⌘⇧F` expands Plugin Settings and focuses its local filter. The open page keeps native selection while the keyboard candidate uses a focus outline until Return opens it.

Plugin search candidate and open page

When the section containing the open page is collapsed, its header uses a separate accent marker instead of looking like another selected row.

Collapsed section containing the open page

### Full Disk Access drag guidance

Permission guidance appears only after the user explicitly requests it.

<p align="center">
  macOS Full Disk Access settings<br>
  MacTools Full Disk Access drag helper
</p>

## Testing

- `make generate`
- Debug and Release builds
- focused XCTest coverage for permission coordination, plugin-host permission routing, Settings navigation and history, sidebar search and preferences, URL routing, MacTools search, and Disk Cleanup's permission declaration
- `make script-tests` (195 tests)
- `make build-plugin PLUGIN=DiskClean`
- manually verified collapsible sidebar sections, distinct open-page/search-candidate/containing-section indicators, dynamic `⌘1…⌘9` navigation, Command-held hint visibility, plugin filtering, explicit permission requests, and Full Disk Access drag guidance
- verified bundled PermissionFlow resources, localization bundles, MIT notice, and strict code-signature validation

Closes #372


ggbond268:
Thanks for the improvements, @xcv58! The extra plugin filter overlaps with the existing in-app search and makes the sidebar feel a bit cluttered. Could we remove the plugin filter and its related functionality for now, while keeping the existing search behavior unchanged?

Longer term, we may refine the in-app search popup to focus on finding and navigating to settings and features, with action execution handled elsewhere. We can revisit that separately.

xcv58:
> The extra plugin filter overlaps with the existing in-app search and makes the sidebar feel a bit cluttered.

Thanks — that makes sense. Since this PR has already been merged, I’ll prepare a separate PR to remove the persistent plugin filter and its related functionality while keeping the global Command Palette unchanged.

I see these as two distinct tools:

- The global Command Palette provides app-wide discovery and navigation across settings, features, and other supported destinations.
- A plugin finder is a lightweight, Plugin Settings–scoped tool for quickly filtering or jumping to one installed plugin.

If we revisit plugin finding later, we could avoid permanent sidebar UI by adding a transient “Find Plugin…” action to the Plugin Settings options menu alongside sorting. It would open an ad hoc search-and-jump interface, navigate to the selected plugin, and then disappear without retaining filter state or changing the sidebar layout.

I’ll keep that possible follow-up separate from the cleanup PR.

PR #376, opened 2026-08-31
fix(window-switcher): include windows from all Spaces

### LINKED ISSUE #357: 窗口切换：多桌面上的窗口无法全部获取
描述：
在桌面1 存在finder窗口A，在桌面2存在finder窗口B和C，在切换窗口UI里，如果当前处于桌面1，能显示窗口A，不显示BC，反之亦然。

预期：
在任何情况下，切换窗口UI里可以获取并展示全部桌面上的窗口

Image

StephaneRavet:
Closes #357

## Summary
- discover eligible layer-zero windows across every macOS Space with public CoreGraphics APIs
- merge CoreGraphics records with Accessibility metadata without ambiguous matching, and use a bounded fallback activation path
- keep per-window shortcuts stable across catalog changes and require explicit reassignment for ambiguous legacy bindings

## Scope
- Window Switcher plugin, focused regressions, README, feature record, and one unreleased plugin changelog
- no PluginKit API, dependency, version, or other-plugin changes

## Validation
- 58 focused `WindowSwitcherPluginTests` passing
- `make build-plugin PLUGIN=WindowSwitcher` passing
- `make script-tests` passing (196 tests)
- PluginKit v5 compatibility check passing
- whitespace checks passing
- manual macOS: with Accessibility granted, the installed Debug overlay listed two Finder `Récents` entries while CoreGraphics reported one as inactive-Space (`isOnScreen == false`, window 6617) and the active peer as on-screen (window 6681)

## Known repository-wide check
`make ci` script/changelog phase passes. Its XCTest phase still reports nine failures outside this diff: `PluginPackageManifestTests` x1, `SystemStatusPluginTests` x4, `DiskCleanPluginTests` x3, and `DeviceBatteryCommandRunnerTests` x1. The focused Window Switcher suite is green.

## Review
- standards/spec: clear
- security: clear
- exact inactive-entry selection remains unavailable through the accessibility driver; discovery is proven in the running overlay and activation paths are regression-tested.

### DIFF
--- a/Plugins/DiskClean/Tests/DiskCleanPluginTests.swift
+++ b/Plugins/DiskClean/Tests/DiskCleanPluginTests.swift
@@ -159,7 +159,10 @@ final class DiskCleanPluginTests: XCTestCase {
         plugin.handleAction(.setDisclosureExpanded(true))
         let trashTitle = try XCTUnwrap(try cleanControl(of: plugin).actionTitle)
         XCTAssertTrue(trashTitle.hasPrefix("移到废纸篓 · 1 项 · 约"), "actual: \(trashTitle)")
-        XCTAssertTrue(trashTitle.contains("GB"), "actual: \(trashTitle)")
+        XCTAssertTrue(
+            trashTitle.hasSuffix("约 \(DiskCleanFormat.bytes(5_368_709_120))"),
+            "actual: \(trashTitle)"
+        )
 
         controller.snapshot = makeScannedSnapshot(
             candidates: candidates,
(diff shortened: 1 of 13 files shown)


ggbond268:
Thanks for improving multi-Space support! Could you please address these two issues before merging?

1. **Keep the window-raising fallback.**  
   `AXFocused` is read-only for many normal windows. Returning immediately when setting it fails skips `AXRaise`, potentially breaking existing window switching. Please still attempt `AXRaise` when focus attributes are unsupported or read-only, while preserving cancellation and timeout checks. Add a regression test for this case.

2. **Allow shortcuts to be reassigned after an app restarts.**  
   Bindings now use the process ID and window ID, which change when the app restarts. The old binding still reserves its key, so the shortcut stops working and assigning it again reports a conflict. Please handle expired window identities so their keys do not remain permanently reserved, without discarding bindings merely because a window is temporarily missing from discovery. Add a restart/reassignment regression test.

Please also manually verify that selecting a window on another Space focuses that specific window—not just its application. Listing the window alone does not confirm successful switching.

PR #377, opened 2026-09-03
Fix System Status process collection

xcv58:
## Summary

- drain command stdout and stderr while subprocesses run so large process lists cannot fill the pipe and stall `ps`
- execute System Status commands asynchronously with timeout termination, exit-status validation, and failure logging
- add regression coverage for output larger than the pipe buffer and timed-out commands

## Root cause

The process collector waited for `ps` to exit before reading its stdout pipe. On systems with many processes, the output exceeded the pipe buffer, leaving `ps` blocked on its write while MacTools waited for termination. The one-second timeout then returned an empty process list, which the UI rendered as “Collecting…” indefinitely.

## Testing

- `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer xcodebuild -project MacTools.xcodeproj -scheme MacTools -configuration Debug -derivedDataPath build/DerivedData test -only-testing:MacToolsTests/SystemStatusSamplerTests -quiet`
- `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer xcodebuild -project MacTools.xcodeproj -scheme MacTools -configuration Debug -derivedDataPath build/DerivedData test -only-testing:MacToolsTests/SystemStatusPluginTests -quiet`

### DIFF
--- a/Plugins/SystemStatus/Sources/SystemStatusCommandRunner.swift
+++ b/Plugins/SystemStatus/Sources/SystemStatusCommandRunner.swift
@@ -1,0 +1,104 @@
+import Darwin
+import Foundation
+
+enum SystemStatusCommandCompletion: Equatable, Sendable {
+    case completed
+    case timedOut
+}
+
+struct SystemStatusCommandResult: Equatable, Sendable {
+    let standardOutput: String
+    let standardError: String
+    let terminationStatus: Int32
+    let completion: SystemStatusCommandCompletion
+}
+
+enum SystemStatusCommandRunner {
+    static func run(
+        path: String,
+        arguments: [String],
+        timeout: TimeInterval
+    ) async -> SystemStatusCommandResult? {
+        let process = Process()
+        process.executableURL = URL(fileURLWithPath: path)
+        process.arguments = arguments
+        process.qualityOfService = .utility
+
+        let outputPipe = Pipe()
+        let errorPipe = Pipe()
+        process.standardOutput = outputPipe
+        process.standardError = errorPipe
+
+        let outputTask = readToEnd(fileHandle: outputPipe.fileHandleForReading)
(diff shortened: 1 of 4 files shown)


ggbond268:
Thanks for fixing this!

A few suggestions before merging:

1. Avoid blocking Swift’s concurrency pool. Task.detached still uses that pool, so synchronous readToEnd() can delay other tasks, including timeout handling. In a 32-command stress test, 50ms timeouts took about 2 seconds. Moving reads to GCD with a continuation reduced this to about 0.2 seconds. Please use nonblocking I/O or GCD and add a concurrent timeout test.
2. Bound the output-draining step. If a child process keeps the pipe open after its parent exits, awaiting the reader tasks can exceed the timeout. Please add a deadline for draining and closing the pipes.
3. Consider safer PID ownership. Sending SIGKILL to a cached PID has a rare PID-reuse risk. This wasn’t reproduced, but the existing PluginProcessGroupLease may provide a useful pattern.

The main concern is the first item; the overall fix looks sound.

xcv58:
Addressed in e483b746.

1. Replaced `Task.detached` + synchronous `readToEnd()` with nonblocking `DispatchSourceRead` drains on a dedicated GCD I/O queue. Added a 32-command concurrent timeout regression test; 50 ms timeouts complete in about 0.34 seconds locally.
2. Added a 200 ms post-leader drain deadline. If inherited pipe descriptors remain open, the runner terminates the owned process group, closes both streams, and returns the output captured before the deadline. The regression case completes in about 0.23 seconds locally.
3. Replaced the cached Foundation `Process` PID kill path with `posix_spawn`, a dedicated process group, and `PluginProcessGroupLease`. The leader remains unreaped while signals may still be sent, preventing PID/process-group reuse during termination.

Validation: `SystemStatusSamplerTests` passed, including the large-output, concurrent-timeout, bounded-drain, and timeout-termination cases; full `make ci` passed.

PR #395, opened 2026-09-07
Remove persistent Plugin Settings sidebar filter

xcv58:
The persistent Plugin Settings filter adds a second search field to the Settings sidebar. This follow-up to #374 removes that field, its query/result state, highlighting, Command-Shift-F shortcut, and filtered-result numbering. Plugin pages remain directly available in the sidebar, with existing sorting, drag ordering, section collapse, and ordinary keyboard navigation preserved.

The global Command Palette continues to provide app-wide discovery, navigation, and its existing supported commands. Command-F still targets the current page's local search. A possible transient Plugin Settings-scoped finder is a separate future change and is not implemented here.

Sidebar focus is restored only for an actual sidebar row click while the Command Palette is closed. Clicking inside the palette or on a view covering the sidebar preserves the current focus. Focus regressions exercise the custom handler without entering AppKit's blocking mouse-tracking loop, and use bounded layout waits with explicit starting focus so they can run without app activation.

Updates the README, localization inventory, regression tests, and changelog, and removes the obsolete filter screenshot. Window coverage also checks that the filter is absent, the existing sidebar search launcher still opens the Command Palette, Command-Shift-F does not dismiss it, and search navigation preserves sidebar geometry.

Context: [maintainer request](https://github.com/ggbond268/MacTools/pull/374#issuecomment-5571280354) and [follow-up scope](https://github.com/ggbond268/MacTools/pull/374#issuecomment-5571720628).

Validation for commit `6c3f8f8a`:
- 128 focused XCTest cases passed with no failures or skips across `AppWindowRouterTests`, `SettingsNavigationCoordinatorTests`, `SettingsSidebarPreferencesStoreTests`, and `MacToolsSearchTests`.
- Changelog validation and `git diff --check` passed.
- Validation used an isolated project with the production app sources on macOS 27 / Xcode beta. The full application test suite, repository script suite, and PluginKit binary compatibility checks were not rerun for this focused update.


xcv58:
Found a potential bug, put PR into draft.

Issue #217, opened 2026-07-24
[feature] Support customizable menu-bar panels (add/remove, component assignment, hide, shortcut, icon, ordering)


ggbond268:
**Title**: Support customizable menu-bar panels (add/remove, component assignment, hide, shortcut, icon, ordering)

## Background / Current State

The menu bar currently has only two fixed, non-configurable panels:

- **Dashboard / component panel** (`PluginComponentPanel`, grid of cards)
- **Feature panel** (`PluginPrimaryPanel`, list)

Panels are hard-wired to click behavior in `MenuBarStatusItemController` and `MenuBarClickBehaviorPreference`: in standard mode, left-click opens the dashboard and right-click opens the feature panel (swappable). Which panel a component/plugin appears in is decided by the protocol the plugin implements, and users cannot change it.

As the number of components grows, users want to group components into their own panels and control panel count, order, visibility, icons, and shortcuts.

## Goal

Introduce a configurable "panel" abstraction so users can customize the set of menu-bar panels.

### Functional Requirements

1. **Default panels**: keep the existing two panels as defaults, and make them **non-deletable**.
2. **Add panels**: allow adding panels, up to a maximum of **5**.
3. **Delete panels**: non-default panels can be deleted; when deleted, their components **fall back to the default panel**.
4. **Component assignment**: allow choosing which panel each component is shown in.
5. **Hide panels**: panels can be hidden (distinct from delete; config is retained).
6. **Shortcut**: each panel can have its own global shortcut.
7. **Icon**: each panel can have its own icon.
8. **Ordering**: panel order is adjustable (defines left→right layout).

### Click Behavior (compatible with today)

- By default, **left-click opens the leftmost panel** and **right-click opens the rightmost panel**.
- The default ordering must match current behavior (left = dashboard, right = feature panel) so existing users see no change.
- With 2–5 panels, left/right map to the leftmost/rightmost panel; middle panels are reached via their own shortcuts or another entry point (interaction TBD).

## Non-Goals / Out of Scope

- **Specific UI design is not decided yet** (settings layout, panel manager interaction, drag-and-drop, etc. to be defined later).
- This issue only organizes requirements; **implementation is deferred**.

## Open Questions

1. **Panel type**: today one panel is a card grid and the other is a list. Do new panels need a "type / rendering mode", or should they be unified into one container that can hold any component? How are components rendered across different forms?
[...]

PR #397, opened 2026-09-07
feat(panels): add in-place layout editing

### LINKED ISSUE #378: feat(panels): add in-place layout editing to Dashboard and Feature Panel
## Summary

Allow users to reorder visible items directly in the Dashboard and Feature Panel through an explicit **Edit Layout** mode.

This should complement the existing layout controls in Settings: the panel provides fast, contextual reordering, while Settings remains the complete place to manage hidden items, reset layouts, and configure plugins.

## Problem

Dashboard and Feature Panel orders are already independent and persisted, but users must leave the surface they are organizing and open Settings to change them. This makes it harder to understand the effect of each move, especially for the Dashboard's mixed-size grid.

Always-on dragging would conflict with the panels' live controls, sliders, buttons, and hover-driven secondary panels, so reordering needs an explicit interaction state.

## Proposed interaction

- Add an **Edit Layout** control to the leading side of the unified panel toolbar when the current surface has at least two visible items.
- While editing, change the control to **Done** and disable normal plugin interactions.
- Dismiss or collapse secondary/expanded content when edit mode begins, and suppress hover-driven secondary panels until it ends.
- Show clear drag handles, drag previews, and insertion or reflow feedback.
- Persist each completed move immediately through the existing display-preference system.
- `Escape` exits edit mode before it dismisses the panel.
- Switching Dashboard/Feature Panel tabs exits edit mode rather than carrying editing state between surfaces.
- If the panel is dismissed, keep already completed moves.

Dragging must not be the only way to reorder. Each item should also expose accessible actions such as **Move Earlier**, **Move Later**, **Move to Beginning**, and **Move to End**, with appropriate VoiceOver announcements.

## Feature Panel behavior

- Present compact reorder proxies containing the plugin icon, title, and drag handle.
- Use an insertion line to show the destination.
- Auto-scroll the list when dragging near its top or bottom edge.
- Collapse expanded details and prevent rows from invoking their normal controls while editing.

## Dashboard behavior

- Keep cards recognizable in their current mixed-span grid, but disable their live content and add a reorder overlay/handle.
- Treat Dashboard order as a linear source sequence, not as arbitrary saved coordinates.
- Convert a grid drop location into a sequence position and show a live preview using the existing deterministic grid-packing behavior.
- Repack the grid after the move without introducing user-managed holes or free-form positioning.

## Relationship to Settings

The first version should reorder **visible items only**. Keep Settings as the full layout-management surface for:

- showing or restoring hidden items;
- resetting an order;
- plugin settings and management;
- Marketplace and uninstall actions;
- managing long lists with search.

Do not add hide/remove actions to the in-panel editor until there is also a clear way to restore those items from the same surface. Changes made in either place should be reflected immediately in the other.

## Implementation direction

- Own a shared edit-mode state at the unified panel level.
- Use surface-specific reorder/drop adapters for the Feature Panel list and Dashboard grid.
- Commit moves through the existing `PluginHost.movePlugin(id:toOffset:on:)` path so the independent surface orders, hidden slots, persistence, and preference backups retain their current semantics.
- Do not require a new plugin API or preference schema for the initial implementation.
- If an item becomes unavailable or is removed during a drag, safely cancel or ignore the stale drop.
- Respect logical ordering in right-to-left layouts and Reduce Motion/increased-contrast preferences.

## Acceptance criteria

- [ ] An Edit Layout control appears on Dashboard and Feature Panel when at least two visible items can be reordered.
- [ ] Entering edit mode disables plugin actions and hover-driven secondary panels.
- [ ] Feature Panel rows can be reordered with insertion feedback and edge auto-scroll.
- [ ] Dashboard cards can be reordered with a live, deterministic mixed-span grid preview.
- [ ] A completed move is persisted immediately and survives panel reopening and app relaunch.
- [ ] Dashboard and Feature Panel orders remain independent.
- [ ] Existing hidden-item positions and preference backup behavior remain intact.
- [ ] Settings immediately reflects moves made in the panel, and vice versa.
- [ ] Done, Escape, tab switching, and panel dismissal have consistent edit-mode behavior.
- [ ] Reordering never accidentally invokes a plugin control.
- [ ] All reorder operations are available by keyboard and VoiceOver without requiring drag and drop.
- [ ] Appropriate move and destination announcements are exposed to assistive technologies.

## Testing

- Unit-test list and grid destination calculations, including first/last positions and mixed card spans.
- Test edit-mode transitions for Done, Escape, tab switching, and panel dismissal.
- Test stale-item handling when the available plugin/component list changes during editing.
- Test persistence, independent surface order, and preservation of hidden slots.
- Manually test scrolling, secondary-panel suppression, mouse/trackpad dragging, keyboard actions, VoiceOver, right-to-left layout, and reduced-motion behavior.

## Out of scope

- Arbitrary free-form Dashboard coordinates, gaps, or card resizing.
- Showing, hiding, adding, or uninstalling plugins from the in-panel editor.
- Replacing the existing layout-management UI in Settings.
- Window snapping or reordering content in external app windows.

## References

- [Apple: Add and customize widgets on Mac](https://support.apple.com/guide/mac-help/add-and-customize-widgets-mchl52be5da5/mac)
- [Apple Human Interface Guidelines: Drag and drop](https://developer.apple.com/design/human-interface-guidelines/drag-and-drop)

xcv58:
## Summary

Reorder visible Dashboard cards and Feature Panel rows directly in the menu-bar panel using whole-card dragging or move menus. Completed moves save immediately, with Undo and Done controls.

- Keep cards and the drop canvas in their committed layout during dragging. Only the insertion marker and destination text preview the move, preventing mixed-size cards from shrinking the drop area beneath a stationary pointer.
- Preserve independent surface orders, hidden-item positions, preference backups, and Settings synchronization. Invalidate stale drags when items or geometry change.
- Support edge scrolling, RTL layouts, localized feedback, and accessibility actions. Escape, tab changes, and panel dismissal end editing while retaining completed moves.

Closes #378. Custom panels and component assignment in #217 remain separate work.

## Validation

- All 11 `PanelLayoutEditorTests` passed after the geometry fix. The new mounted-view regression covers both surfaces and layout directions, stable card frames and canvas bounds during preview and leave, and reordering after commit.
- A standalone mounted production-view check confirmed that the mixed-size Dashboard canvas stays at 420pt during dragging; the original implementation shrank it to 318pt.
- Changelog validation and `git diff --check` passed. No full application test suite was run for this focused fix.
- The preceding PR revision passed GitHub Build and test. The standalone native drag fixture failed locally on both the original implementation and the stable-layout control; full native drag acceptance is therefore still unverified locally.

Physical mouse/trackpad dragging, sustained edge scrolling, keyboard navigation, VoiceOver, and Reduce Motion require manual acceptance on supported macOS versions. See `docs/testing/panel-layout-editing.md` for fixture boundaries and checks.


ggbond268:
Thanks for working on this! Reordering panel items isn’t a frequent task, and users can already do it in Settings. Adding in-panel reordering introduces more complexity to both the panel UI and the code, so I’d prefer not to support it for now.

I’m also planning to implement custom panels in #217, which may change how we approach panel organization and reordering. I’ll keep this PR on hold—we may revisit it as those needs become clearer.

xcv58:
Thanks for explaining your concerns. I understand wanting to keep the panel UI simple and avoid maintaining something that #217 may replace.

However, I don’t think infrequent use is a strong reason to keep reordering exclusively in Settings. Frequency should determine how prominent the control is; the task itself should determine where it works best. Reordering is a visual task, especially for Dashboard cards with different sizes. A Settings list supports changing the order, but doesn’t provide the same experience as seeing the resulting arrangement while editing it.

We could address this in a few ways:

- Put **Customize Layout…** in a context or overflow menu, showing editing controls only when explicitly requested.
- Add a full, interactive panel preview to Settings.
- As a smaller step, keep the existing Settings reorder list and add a live preview beside it.

The preview options could also fit naturally into #217’s broader panel customization work. An interactive preview would still carry some editing complexity, though, so moving it into Settings wouldn’t necessarily eliminate that cost.

I’m open to holding or simplifying this particular implementation. But I’d push back on treating the existing Settings list as a sufficient substitute. Could we agree that users should be able to arrange items while seeing the actual resulting layout, and then choose the implementation that best fits #217?

ggbond268:
Thanks for the thoughtful feedback. I agree that arranging items while seeing the actual layout is more intuitive.

After thinking it over, I’d prefer to keep layout preview and reordering together in the panel, and remove the duplicate reordering controls from Settings. The UI and interaction details still deserve some careful thought, but we can refine them as part of the broader panel customization work.

I’m happy to move forward with merging this PR. I’m currently working on custom panels in #217 and will build on your implementation with further improvements.

xcv58:
Thanks for reconsidering! Keeping reordering in the panel and removing the duplicate controls from Settings sounds like a good direction. It gives users one clear place to arrange items while seeing the result.

I’m happy for #397 to serve as the foundation for your custom-panel work in #217, with further UI refinements happening there.

Would you prefer me to remove the Settings reordering controls in this PR before merging, or leave that cleanup for your follow-up work?

PR #406, opened 2026-09-09
feat(mouse-enhancer): add separate mouse and trackpad scroll tuning

### LINKED ISSUE #400: feat(mouse-enhancer): add adjustable scroll step size and speed gain
## Summary

Mouse Enhancer currently covers scroll *direction* (vertical/horizontal inversion for mouse and trackpad/Magic Mouse) and middle click, but not scroll *feel*: there is no way to tune how far each wheel tick scrolls or how sensitive the response is. I'd like to propose adding scroll tuning options — at minimum a step size (minimum distance per wheel tick) and a speed gain (multiplier applied to scroll deltas), ideally per device.

## Motivation

Mice vary a lot in what they report per wheel tick:

- High-resolution / free-spin wheels emit many small deltas per physical tick, so pages can feel like they are flying away with only a light flick.
- Some mice and wireless receivers report tiny deltas, so every tick barely moves the page and users end up spinning the wheel constantly to get anywhere.

macOS offers no help here: its only relevant control is natural scrolling, which is global (trackpad included), and there is no system-level step or gain control for mice. So today a user whose wheel feels wrong can only invert direction — they cannot slow it down or speed it up.

## Proposed behavior

- **Scroll step size**: minimum distance a single wheel tick produces.
- **Speed gain**: multiplier applied to the reported scroll deltas.
- Per-device settings if the event pipeline already distinguishes mouse vs. trackpad (the inversion settings already do), so tuning a mouse does not change trackpad feel.
- Defaults that match today's behavior, so nothing changes until the user tunes it.

## Reference

[Mos](https://github.com/Caldis/Mos) handles this area really well: alongside reverse scrolling it exposes scroll step and speed/duration (plus optional smoothing), which is exactly the missing "feel" layer. Its approach of intercepting wheel events and re-emitting adjusted deltas could be a useful reference — Mouse Enhancer's event tap could apply step/gain in the same pass as inversion.

icy37785:
Closes #400

## Summary

Add separate scroll step and speed controls for mice and trackpads, so small wheel ticks can be raised to a minimum distance and scrolling can be slowed down or sped up.

- Step: 0–120 px per event; speed: 0.1×–5×. Defaults preserve existing scrolling.
- Apply speed and the pixel floor with one proportional scale, rounding integer deltas only at the final conversion. Combining the controls preserves consistent line, pixel, and fixed-point distances, subject to integer rounding. Integer overflow is clamped safely.
- Normalize tuning values when loading saved settings and when editing them. Non-finite values use defaults; finite values are constrained to supported ranges and increments.
- Keep independent direction reversal and bypass recognized continuous events from remote-control software. All controls use the existing event tap and localized settings UI.

## Validation

- 41 MouseEnhancerPluginTests passed in an isolated SwiftPM harness using the plugin sources, including 8 added regression tests for combined tuning, rounding, integer limits, trackpad reversal, and invalid saved settings.
- The isolated UserDefaults NaN crash reproduction now exits successfully.
- Changelog validation and git diff --check passed. The full app test suite was not run.


icy37785:
Heads up on CI: the first run failed in `ClipboardPasteboardReaderProcessTests.testPlainTextRequestRejectsSensitiveProducerTypesAndRecovers`, which is unrelated to this PR's Mouse Enhancer diff. The same base commit passed "Build and test" on `main` earlier the same day, and the test class passes locally (12/12). It looks like a flaky process-spawn test under CI runner load — a re-run should clear it. (A second run then caught an over-long changelog entry from my side; fixed in 06887a37.)

PR #410, opened 2026-09-10
Refine Clipboard UI, opening performance, and source attribution

ggbond268:
## Summary

Refine the Clipboard workspace so common actions are easier to find, previews stay readable, and reopening feels more immediate. This also fixes source attribution for Universal Clipboard content, which could previously appear to come from whichever local app happened to be in front.

@xcv58, as the original author of the Clipboard workspace in #335, please take a look at the interaction, preview, and capture changes below.

## Settings

- Organize settings into **Clipboard History**, **Snippets**, **Paste Queue**, **Advanced**, and **Local Data**. Put **Privacy & Storage** first in Clipboard History and make **Enable Clipboard History** the collection control.
- Move infrequent retention, collection-shortcut, snippet, and queue options into their feature's Advanced disclosure. Keep excluded apps, private-copy shortcuts, and clipboard-window shortcuts together in the shared Advanced section. Settings search still reveals the relevant collapsed shortcut group.
- Replace repetitive headings and the privacy-details disclosure with concise rows. History, saved-clip, and snippet cleanup actions remain directly visible, retain their confirmations, and have enough width to display their complete button labels.
- Present encrypted backup, restore, and local rollback alongside Local Data cleanup, using the same settings rows, neutral icons, and short buttons with complete labels. Preserve file selection, password entry, replacement confirmation, keyword-capacity confirmation, and recovery behavior.
- Extract `PluginSettingsItem` into PluginKit so the Clipboard settings and host form rows share the same icon, title, description, and trailing-control layout. Require host 1.3.1 for this newly exported type, document that requirement, and add it to the compatibility inventory. Host 1.3.0 has already shipped without this API, so the Clipboard manifest now requires 1.3.1 and the development marketing version predeclares that compatible host; build numbers, plugin package versions, and published catalogs are unchanged.
- Preserve collection, encrypted storage, setup guidance, exclusions, retention, snippets, keyword expansion, paste queues, and shortcut configuration while reorganizing their presentation. **Manage Snippets** selects its tab immediately and preserves its requested destination through cold preparation, joins already-running preparation, and reuses a warm page. Empty libraries still open the Snippets view; a later ordinary opening retains its normal default scope.

## Clipboard panel and paste queue

- Use **All / History / Snippets** as the primary scope switcher, with secondary content filters in a compact menu. Keep selection controls aligned and center each row's icon vertically.
- Widen the list to a **40/60 list-to-preview split** and align the rich-text preview with the detail header.
- Remove the top pause/close controls. Group Settings, the persistent collection toggle, and keyboard hints on the left of the footer, with one divider between buttons and hints. Remove redundant footer information.
- Give footer buttons subtle hover feedback. Use a compact crossed-clipboard icon for the collection toggle, retaining its full click target; disabled collection is highlighted with the system accent color, and the tooltip describes the available action.
- Apply the same restrained frosted surface to History, Actions, and the paste-queue HUD. A mostly opaque semantic backdrop improves readability over busy desktops, and Reduce Transparency uses an opaque fallback.

## Window placement

- Open on the display under the pointer, centered in its available work area until the user moves the panel.
- Remember a separate position for each display across launches. Store display-relative offsets and clamp restored frames to the current work area, including display rearrangements and smaller screens.
- Record explicit window dragging and live resizing without allowing programmatic centering/restoration or automatic display relocation to overwrite remembered positions. Use the existing host display-topology notifications to refresh the geometry baseline. A subsequent user move is saved on its current display; disconnecting and reconnecting an external display preserves each display's independent position. Reconfiguration during a drag ends that tracking session until the next deliberate move.
- Integrate the snapping guides merged in #384 with per-display placement. The default centered frame stays aligned with the snap reference, and position persistence waits for the final snap before saving. Closing, reopening, or a display reconfiguration cancels active guides; display changes end movement tracking even when the native move notification arrives before the host refresh.

## Rich-text appearance and reuse

- Follow the app appearance by default and allow a local light/dark preview override. Prepare both appearances so switching does not repeat document import.
- Adapt unreadable foreground and decoration colors for the selected canvas while preserving readable colors, explicit highlights, fonts, and links. Original clipboard representations remain unchanged.
- Keep HTML import on the main actor as required by AppKit; perform eligible payload reads and RTF preparation off the main actor. Preserve bounded imports, fallback text, retry behavior, and external-resource restrictions.
- Retain only the most recent bounded prepared rich-text preview across ordinary closes. Changed/deleted content, explicit discard, deactivation, and memory pressure invalidate it; failures and cancelled imports do not populate the cache, and source payloads are released.

## Opening performance

- Prepare the recent-items page after history and snippet metadata load, before the first presentation. Suspend preparation during backup operations and prepare again after restored storage has loaded; the snippet manager also respects the backup action lock. Opening during preparation joins the pending work.
- Remove duplicate metadata counting and unnecessary per-scope sorting, reuse identifier sets, start background preparation before native window construction, skip unchanged frame updates, and disable the panel opening transition.
- Add a synthetic native-window benchmark and document the measurement limits. With 10,000 synthetic records in Debug, the recorded first-open baseline was approximately 600 ms; updated prepared first opens had a 227 ms median and reopens a 21 ms median. The baseline was a single exploratory run, so this is indicative rather than a controlled release benchmark. These measurements exclude global shortcut dispatch, physical display latency, and asynchronous preview completion.
- Preserve pagination, revision/cancellation checks, bounded image/PDF caching, asynchronous persistence, and serial background OCR. Opening the panel does not trigger OCR.

## Source attribution and compatibility

- Recognize `com.apple.is-remote-clipboard` as **Other Device**, with **Universal Clipboard** in details, instead of attributing that content to the foreground app. Remote provenance takes precedence over application hints.
- Respect valid `org.nspasteboard.source` declarations. Empty, invalid, unreadable, or conflicting declarations remain **Unknown Source**; copies without a marker retain the existing foreground-app estimate.
- Read source declarations inside the isolated pasteboard helper and carry the result alongside the captured payload. Source markers are advisory: existing foreground/recent-app privacy checks remain in force, and a declared app is also checked against exclusions.
- Preserve source metadata through saves, recapture, encrypted lazy reloads, and backup restoration. Matching-ID backup merges retain known remote provenance even when the other record uses the legacy format, and remote records cannot inherit an unrelated local application. Existing application/unknown records and archives remain readable without a format migration or rewriting old history.
- This recognizes available pasteboard metadata; it does not infer an exact remote device/app or claim to identify every AirDrop transfer.
- Update copy across all 11 supported locales and retain the manifest-only `settings.storage.encrypted.description` key required for plugin package synchronization.

## Validation

- Built the app and plugin targets after integrating #384. All **26 focused window-placement, movement-tracking, snap-geometry, and overlay tests** passed. Added two regression tests for matching default/snap frames across displays and window sizes, and for recording the final snapped frame.
- Ran `make ci`: all **254 script tests** and changelog validation passed. XCTest ran **4,588 cases: 4,586 passed and two failed**; all **713 Clipboard cases passed**, including the previous review fixes and the final display-notification-order handling.
- Rechecked both XCTest failures independently. `ActionGridOverlayControllerTests.testConfirmationRequiredActionUsesGridAnchoredPresenter` passed. `AppleShortcutsCommandRunnerTests.testTimeoutForceKillsTermIgnoringDescendantAndDrainsPipes` still failed because its `child.pid` fixture was missing, matching the previously observed local failure. Both areas are unchanged from main.
- Ran the remaining CI steps separately after the XCTest failure stopped `make ci`. **PluginKit v6 binary compatibility passed**. The native panel-layout fixture failed all three built-in dashboard/LTR attempts because the simulated drag was cancelled instead of saved; the fixture and every production source it compiles are unchanged from main. Therefore the complete local CI run is not green; no unrelated workaround is included in this PR.
- Recompiled an isolated in-memory probe against the final plugin binary. It exercised the real Clipboard drag callback, release detection, final snapping, persisted position, reopening, and automatic relocation; all assertions passed. It did not read the system clipboard or install the app.
- Retain the nine earlier review regression tests for cold/warm/in-progress Snippets presentation and per-display movement tracking. At the preceding `eaffe26b` head, the real-controller Snippets probe passed cold, in-progress, and warm openings, and GitHub's full Build and test workflow passed.
- The source-attribution fixtures publish complete `NSPasteboardItem` representations using `clearContents` / `writeObjects`, assert revision advancement, and preserve source-priority and privacy checks through the real helper process. These tests passed in the current full run.
- `git diff --check` passed. Display-reconfiguration behavior is covered with synthetic screen geometry; physical display hot-plugging was not exercised. The snapping integration consumes the API already merged in #384 and adds no further PluginKit API or minimum-host increase. GitHub CI will rerun for this updated head.

## Scope

This branch contains only Clipboard changes, their tests/documentation/benchmarks, and the shared settings-row extraction needed by the Clipboard settings. The mixed `README.md` and `SettingsView.swift` edits were selected by hunk; the separate custom menu-bar panel and menu-bar quit-icon work is excluded.


xcv58:
Thanks for this update—the Clipboard improvements look useful. I’ve manually tested most of the changes, and everything I tested passed. A standalone review of `e1eefa7f` identified three follow-ups:

**1. “Manage Snippets” can lose its requested destination on a cold opening**

The new Settings callback calls `showSnippets()`, which starts asynchronous preparation and immediately selects Snippets. When preparation completes, `applyPreparedPresentation()` resets the scope to History/All.

This reproduced through the actual panel controller with synthetic data: the cold opening landed in History, while the prepared opening correctly stayed in Snippets.

Could we preserve the requested destination through preparation and add coverage for cold, preparation-in-progress, and warm openings?

**2. Position saving stops after display geometry changes while the panel remains open**

`rememberPanelPosition()` requires the current display configuration to equal the snapshot captured when the panel opened. That snapshot isn’t refreshed by the display-change callbacks.

After connecting/disconnecting a display or changing resolution, subsequent deliberate moves appear to be ignored until the panel is presented again. This is a source-level finding; physical reproduction is still pending.

Could we refresh the geometry baseline after display changes while continuing to distinguish automatic repositioning from a user’s next deliberate move?

**3. Two source-attribution tests need a fresh pasteboard revision**

Both of these tests failed in focused and full local runs:

- `testHelperReturnsSourceHintsSeparatelyFromClipboardPayload`
- `testConflictingDeclaredSourcesAreUnknownAndPrivateMarkersStillBlockCapture`

They add a marker after the helper has already read the pasteboard. An independent probe confirmed that `setData` succeeds without advancing `changeCount`, allowing the helper to retain cached type information.

Could we construct complete `NSPasteboardItem` representations and publish a fresh revision with `clearContents` / `writeObjects` before each read? The source-priority and privacy assertions should remain unchanged. This finding concerns test publication, not a demonstrated production privacy bypass.

For context, the unchanged PR head passed **4,522/4,524 XCTest cases**, all **254 script tests**, and PluginKit binary compatibility. GitHub CI is green; the two local failures are the fixtures above.

PR #412, opened 2026-09-10
feat(screenshot): add native capture and editing plugin

idev19:
Adds Screenshot as a native MacTools plugin, adapted from [Snap](https://github.com/idev19/Snap), with region/window capture, quick copy, annotations, local OCR and QR recognition, pinned images, scrolling screenshots, and silent region recording on macOS 15 or later. Other features support macOS 14.

The plugin uses host permission cards, declarative settings, optional shortcuts, and foreground interactive actions. Disabling it closes its windows and cancels pending work. Processing stays local; exported files are preserved on uninstall.

- QR masking applies an opaque fill in preview and PNG output, including after later blur or mosaic edits.
- Scrolling capture matches, copies, and composes on a serial background actor. Appended strips own their pixels; output is capped at 128 MiB of RGBA pixels and 32,768 rows. Finish includes accepted frames, and cancellation suppresses late results.
- Adds plugin packaging, localized resources, documentation, and runtime action coverage. Control-window tests use real AppKit window IDs.

Validation:
- All 48 native Screenshot XCTest methods passed in an isolated SwiftPM harness, including QR decoding at 1×/2×, frame-provider release, background pixel processing, output limits, and finish/cancel ordering.
- The native Xcode host build and selected Screenshot test classes plus `PluginRuntimeActionSnapshotTests` passed with signing disabled.
- All 254 repository script tests, changelog validation, and `git diff --check` passed. The full application XCTest suite was not run.

Real screen capture/recording, mixed-scale display interactions, and system permission prompts still need manual validation; see the [Screenshot guide](docs/plugins/screenshot.md#development-and-validation).


ggbond268:
Thanks for bringing Screenshot to MacTools! I reviewed `24680d6d` and found four issues worth fixing before merging:

1. **Make QR masking reliably hide the content.** A generated 620×620 QR code remained fully readable after applying the current mask and exporting the PNG. Fixed-size pixelation can preserve the QR pattern, so sharing the image could expose information the user intended to hide. Please use an opaque fill for the “Mask QR” action and add a test that verifies the exported image no longer decodes to the original payload.

2. **Avoid retaining full frames during scrolling capture.** `CGImage.cropping(to:)` retains the original image. In a synthetic test, 20 frames scrolling by four rows produced about 5 MB of output pixels but kept about 92 MB of source pixel buffers alive. Please copy each appended strip into an independent buffer and add an output-size or memory limit.

3. **Move stitching off the main thread.** `ScrollCapture.tick()` runs the matching algorithm on the main actor every 250 ms. With optimization enabled, matching alone took about 25 ms for a 2160-pixel-high frame and 69 ms for a 3600-pixel-high frame. Please move matching and composition to a serial background worker, preserve cancellation and result ordering, and consider reducing the search cost.

4. **Use real window identities in the tests.** On my machine, 38 of the 40 native XCTest methods passed. The two control-window tests raised an AppKit exception during initialization because `CaptureTestWindow.windowNumber` returns arbitrary IDs such as 101. Please use actual window IDs, or extract the ID-validation logic so it can be tested without overriding AppKit's window identity.

These checks used synthetic images and the plugin's native XCTest suite in an isolated SwiftPM harness. Real capture/recording and mixed-scale display checks are still pending. The QR masking issue is the main privacy concern.

PR #413, opened 2026-09-10
feat(DeviceBattery): add JBL headphone battery support

AndrewDi:
- Add JBLHeadphoneSupport.swift for JBL device detection and BLE advertisement parsing
- Add .jblHeadphone kind to DeviceBatteryKind with headphones icon
- Support split battery (left/right/case) for JBL Tour Pro, Live Pro, Quantum, Sense Lite models
- Detect JBL headphones by name/manufacturer in both GATT and advertisement scanning
- Use headphones SF Symbol for all JBL headphone variants
- Add jbl/headphone/earbuds keywords to plugin discovery

### DIFF
--- a/Plugins/DeviceBattery/Sources/DeviceBatteryComponentView.swift
+++ b/Plugins/DeviceBattery/Sources/DeviceBatteryComponentView.swift
@@ -723,6 +723,21 @@ func deviceSymbolName(for item: DeviceBatteryItem) -> String {
         return "computermouse.fill"
     case .airPodsPart:
         let ownName = item.name.lowercased()
+        
+        // Check for JBL headphones first
+        if containsAny(haystack, ["jbl"]) {
+            if ownName.contains("case") || ownName.contains("充电盒") {
+                return "headphones"
+            }
+            if ownName.contains("左耳") || ownName.contains("left") || ownName.contains("🄻") {
+                return "headphones"
+            }
+            if ownName.contains("右耳") || ownName.contains("right") || ownName.contains("🅁") {
+                return "headphones"
+            }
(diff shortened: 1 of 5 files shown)


AndrewDi:
@ggbond268 PTAL

ggbond268:
Thanks for adding JBL support! I found a few issues that need addressing before merging:

1. **The plugin does not compile.**  
   A plugin-only build fails because existing switches do not handle `.jblHeadphone`. The new JBL readings are also assigned to a dictionary that expects Apple-specific readings. Please update the affected switches and use a shared battery-reading type where appropriate.

2. **The advertisement format needs verification.**  
   The parser uses company ID `0x0001`, which is assigned to Nokia in the [Bluetooth SIG registry](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Assigned_Numbers/out/en/index-en.html). The battery offsets and charging flags also need validation against real JBL packets. Please confirm the format for supported models and add captured-packet tests; changing the company ID alone would not establish that the parser is correct.

3. **Advertisement readings are not connected to the existing device flow.**  
   The scan plan currently enables manufacturer-advertisement scanning only for Apple headphones. Additionally, JBL readings are stored under a new `jbl-UUID` ID, but the output step only accepts IDs from the original target list, so those readings would be discarded. Please include verified JBL targets in the scan plan and associate readings with an existing connected device, preserving its identity and skipping ambiguous matches.

I’d suggest reusing the existing macOS and standard GATT battery paths where supported, and deferring private advertisement parsing until it is verified on hardware. Please retain the existing scan limits and add focused tests for JBL parsing, device matching, and AirPods regressions.

Issue #417, opened 2026-09-10
CLI: offer optional managed installation in stable releases


xcv58:
## Proposal

Offer the existing MacTools CLI as an optional, separately downloaded tool in stable releases after the remaining signed Nightly validation is complete. #409 delivered managed Nightly installation; #403 explicitly excludes stable installation, so this issue requests agreement on that next scope. Related RFC: #309.

@ggbond268, could you review whether this limited stable rollout is the right next step, and confirm the supported platforms and compatibility expectations before implementation?

## Proposed first release

- Keep the current command surface: help, version, doctor, action discovery/description/availability, and execution of currently eligible parameterless actions.
- Provide Install CLI in the stable app's settings, with explicit confirmation and a separate signed/notarized download. Never install automatically for users who have not opted in.
- Manage updates with the app and support removal and explicit rollback. Preserve normal macOS background permission and quarantine checks.
- Keep stable and Nightly commands, services, installation roots, receipts, and update state isolated. Preserve manual and Homebrew installations.
- Start with Apple silicon, subject to maintainer agreement; agree on the minimum macOS version and verify every advertised platform.
- Document command, JSON, exit-code, and protocol compatibility before making stable guarantees.

## Validation so far

A pinned signed Nightly candidate passed fresh installation and uncached notarization-ticket recovery in clean macOS 26.6.2 and 27.0 RC VMs. Both systems automatically enabled initial background registration. Native permission denial, pending-approval guidance, the Settings link, and recovery without reinstallation passed on both. An OS-generated first-registration pending-approval prompt was not observed.

Physical-environment checks also covered installation, integration toggles, removal, regular-file collisions, and selected update, rollback, offline, and migration scenarios. Coverage and evidence provenance are recorded separately in the accompanying documentation PR. The reports omit personal machine names, paths, distribution addresses, and signing identities. The disposable VMs and raw VM evidence were removed after validation; the report retains the result summary.

## Release requirements

- [ ] Resolve the remaining signed acceptance checks tracked under #403, including the outstanding OS-specific update/migration/rollback cases, Sparkle update path, and collision/isolation/failure matrix.
- [ ] Agree on supported architectures and macOS versions; validate the oldest supported OS.
- [ ] Validate stable-specific signed artifacts, managed installation, and app/CLI update compatibility.
- [ ] Validate stable/Nightly coexistence and manual/Homebrew collision handling.
- [ ] Review compatibility contracts, diagnostics, localization, and recovery instructions.
[...]

Issue #403, opened 2026-09-08
CLI: add in-app installation and managed lockstep updates


xcv58:
# Summary

Add an in-app, one-confirmation workflow that downloads, verifies, installs, updates, and removes the separately distributed Nightly CLI. This preserves the separate CLI artifact and app download size established in #309 while removing the current manual ZIP and PATH setup for most testers.

The first implementation should remain Nightly-only and Apple-silicon-only.

# User experience

In **Settings → General → Command Line**, show one of these states:

- Not installed
- Downloading / Verifying / Installing
- Installed, including CLI version and path
- Update available
- Update failed or app/CLI incompatible

For an uninstalled CLI, show **Install CLI…**. The confirmation must explain:

- The matching Nightly CLI version and download size
- The managed installation directory and `mactools-nightly` command path
- That MacTools will not overwrite an existing unmanaged command
- Whether Command-Line Integration will be enabled
- That a managed CLI updates automatically with MacTools; there is no separate update switch

The normal case should require one MacTools confirmation. macOS may still require separate Login Item or background-item approval, which the app must explain and link to rather than bypass.

After installation, expose **Update**, **Remove**, **Reveal in Finder**, and **Copy CLI Path** actions as appropriate.

# Ownership and installation layout

- Install the executable in a versioned app-managed directory, for example `~/Library/Application Support/MacTools Nightly/CLI/<version-build>/mactools`.
- Create `~/.local/bin/mactools-nightly` as an atomic symlink to the managed executable.
- Record a managed-install receipt containing the channel, version, build, source release, executable hash, managed path, and link path.
- Refuse every collision at the command destination, including files, directories, and dangling symlinks, unless the existing link is provably owned by this MacTools installation.
- Never modify a CLI installed manually, by Homebrew, by another MacTools channel, or at an unmanaged path.
- Never install a CLI merely because the app was installed or updated.
- Do not modify shell startup files automatically. Show copyable PATH guidance when `~/.local/bin` is absent from the user PATH.

# Release metadata and verification

Publish machine-readable metadata that binds the CLI to the matching Nightly app build. It should include at least:

- Immutable asset URL
- SHA-256 digest and size
- App version and build
- CLI version and build
- Architecture
- Expected signing identifier and Team ID
- Release channel and source commit

The app must not scrape GitHub HTML or rely on human-readable release notes. Before installation it must:

1. Download into a unique staging directory with bounded size and timeout.
[...]

PR #418, opened 2026-09-10
docs(cli): propose optional stable distribution and record signed validation

xcv58:
Managed CLI installation is currently Nightly-only. This PR proposes offering the existing limited CLI as an optional stable download, with platform, compatibility, packaging, and coexistence requirements that must be met before publication.

It also records signed Nightly validation on physical environments and clean macOS 26.6.2 / 27.0 RC VMs. The report distinguishes completed checks, user-reported results, and outstanding release checks. Initial background registration was automatically enabled; permission denial and recovery passed, while an initial OS-generated pending-approval prompt was not observed. Personal identifiers are omitted, and raw VM evidence removal is documented.

Proposal discussion: #417. Related: #403, #409, and #309. This documentation PR leaves #403 open and does not enable stable publication.

@ggbond268, please review the proposed stable scope and release requirements before we begin the implementation PR.

Validation: checked both documents for personal machine/account/path identifiers, verified relative links, and ran `git diff --cached --check`. Only two Markdown files change; no product code or release settings change, so builds and runtime tests were not rerun.

### DIFF
--- a/docs/cli/proposals/stable-cli-v1.md
+++ b/docs/cli/proposals/stable-cli-v1.md
@@ -1,0 +1,83 @@
+# Proposal: an optional stable MacTools CLI v1
+
+Status: draft for maintainer discussion; stable publication is not enabled. See the [signed Nightly acceptance report](../validation/2026-09-10-nightly-signed-acceptance.md) for completed checks and remaining gaps.
+
+Proposal discussion: [#417](https://github.com/ggbond268/MacTools/issues/417).
+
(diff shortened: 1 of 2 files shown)


ggbond268:
Sorry, I merged this PR by mistake and have since reverted it. Could you please open a new PR so we can continue from there?

I agree with the proposed direction: distribute the CLI as a separate package and provide an Install CLI button in General settings, so users can download and install it when needed. Please proceed on that basis, with the remaining validation completed before enabling stable publication.

Thanks for your work, and sorry for the extra step!

Issue #416, opened 2026-09-10
Window Switcher: add searchable window navigation


xcv58:
## Problem

The current chooser expands every window into a large app-icon grid. With many Chrome windows, icons repeat, titles become unreadable, and the panel overflows the display. The default persistent letter grid also replaces native Command-Tab without preserving hold/cycle/release behavior.

Parent roadmap: #414

Depends on: #415 for stable destinations, window recency, cached discovery, and action outcomes. UI prototyping can proceed independently, but release acceptance requires that foundation.

## Source evidence

Inspected revision: [835a78a](https://github.com/ggbond268/MacTools/tree/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher).

- [The overlay sizes to its content](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherOverlayController.swift#L278) and [renders a fixed-size grid without scrolling](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherOverlayController.swift#L524).
- [Default mode is the persistent key grid](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherModels.swift#L50); release-to-switch exists only in direct-cycle mode.
- [Key-grid input resolves letter/Command-letter tokens](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherOverlayController.swift#L363), with no search field or normal selection navigation. The tile's quit action targets the whole app.

## Proposed experience

| Intent | Recommended preset |
| --- | --- |
| Switch apps | Preserve actual native Command-Tab |
| Switch current-app windows | Enhanced Command-backtick |
| Switch all individual windows | Option-Tab |
| Find a particular window | Persistent search entered from the chooser or a configurable shortcut/canonical action |

Allow an intentional Command-Tab replacement configuration. Preserve existing bindings until the user explicitly selects a new preset; explain how old per-window letter assignments are migrated or retired. Check conflicts through the existing shortcut system.

### Chooser

[...]

Issue #415, opened 2026-09-10
Window Switcher: fix window tracking and activation reliability


xcv58:
## Problem

Window Switcher must reliably identify, order, and activate individual windows before a richer chooser can be dependable. With many Chrome windows, array position and app-level recency are insufficient to identify the user's intended destination.

Parent roadmap: #414

## Source findings

Inspected MacTools revision: [835a78a](https://github.com/ggbond268/MacTools/tree/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher). Revalidate against the implementation branch before fixing; these observations do not identify the installed screenshot build.

1. [Window IDs use process ID plus array index](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherAppCatalog.swift#L210). Windows are sorted by minimized state, position, and title, so moving/closing windows can change the meaning of an ID.
2. [Shortcut identities use application occurrence order](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherModels.swift#L493), allowing assignments to follow a position instead of a particular window.
3. [Recency is maintained for applications](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherAppCatalog.swift#L89). [Initial cycling anchors to the first matching app entry](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherPlugin.swift#L544), not the focused window.
4. Discovery runs synchronously on the main actor when invoked. Workspace app activation/launch/termination notifications do not provide complete window lifecycle or intra-app focus tracking.
5. [Activation discards AX action results](https://github.com/ggbond268/MacTools/blob/835a78a315b367fcaa7d86402dd414fad9fcb092/Plugins/WindowSwitcher/Sources/WindowSwitcherAppCatalog.swift#L146); a stale or unresponsive destination can fail without a meaningful outcome.
6. AX enumeration failure is collapsed to an empty list, which can become an app-only fallback. Distinguish unavailable data from a confirmed windowless app.
[...]

Issue #414, opened 2026-09-10
Window Switcher: redesign goals and roadmap


xcv58:
## Problem

The Window Switcher becomes difficult to use when one application has many windows. A user-provided screenshot shows Chrome dominating an icon grid, repeated app-name labels, heavily truncated titles, and a panel extending beyond the usable screen. This defeats the purpose of quickly finding a specific window and makes unrelated applications harder to reach.

The screenshot establishes an identification and layout problem. It does **not** prove duplicate enumeration or incorrect activation: distinct real windows can share the same title. The original image is not attached because some personal window titles remain visible; a redacted copy can be added separately.

## Agreed product direction

Make finding and activating an existing window fast and predictable while preserving native macOS app-switching simplicity. Use native Command-Tab as the baseline: one entry per app and familiar hold, cycle, release behavior.

Recommended preset:

| Intent | Interaction |
| --- | --- |
| Switch applications | Preserve the real native Command-Tab |
| Switch windows of the current application | Enhanced Command-backtick window chooser |
| Switch individual windows across applications | Option-Tab window chooser |
| Find a particular window | Persistent app/title search entered from the chooser or a configurable shortcut/canonical action |

Bindings remain configurable. Existing user choices must not be silently overwritten. Offer the native-compatible arrangement as an explicit migration preset, and allow an intentional Command-Tab replacement configuration.

Quick cycling and search have distinct input states. Releasing the modifier commits during cycling; entering search keeps the chooser open after modifier release so ordinary text editing and Chinese IME composition work. Cancel must preserve the original foreground target.

## Short-term goal: first dependable release

- Stable live-window identities and actual per-window recency, including focus changes within one app.
- Correct target activation, safe handling of stale windows, and separate window-close/application-quit actions.
- Cached state and asynchronous discovery so a slow application cannot block opening the chooser.
- Current-app and all-window scopes with a bounded, scrollable list, readable titles, clear selection, and automatic selection scrolling.
- App/title search with proper keyboard navigation and Chinese text input.
- Optional preview of the selected window, loaded asynchronously. Core switching remains usable without Screen Recording permission.
- Explicit display/Space scope and minimized/hidden state handling; test actual platform behavior before claiming coverage.

## Implementation tracking

- [ ] #415
- [ ] #416

#416 depends on #415 for stable destinations, actual window recency, cached discovery, and reliable action outcomes. UI prototyping can proceed independently; release acceptance requires both.

## Long-term goal: polished window navigation

[...]

PR #419, opened 2026-09-11
Redesign Window Switcher with searchable window navigation

xcv58:
Window Switcher now supports searching by window title or app, browsing a grid or list, and inspecting a preview before switching to an exact window. Real windows on other desktop or fullscreen Spaces remain discoverable even when macOS omits them from the Accessibility window list.

## Behavior

- **Search and Select** is the default for new installations. Existing profiles retain **Direct Keys** and saved assignments. Switching modes is explicit; existing keys never silently become Close Window or Quit App commands.
- **Cycle Windows** switches on modifier release. Deliberately typing or focusing Search keeps the chooser open until Return or dismissal, without a timed mode transition.
- The chooser sizes itself to the window count, layout, and preview setting. Searching keeps its frame steady. Explicit scope, display, layout, and preview changes resize it while preserving the top edge within screen bounds. Search and Select retains manual sizing per layout/preview combination during the controller lifetime; **Reset Size** restores automatic sizing.
- Grid and list support search, visible-result number shortcuts, editable direct-key badges, and clickable overflow indicators. The current-app shortcut narrows to the highlighted app. Scope controls share Clipboard History's flat grouped appearance, hiding a redundant app choice unless that filter is active.
- Right-click, Control-click, or Shift-F10 exposes window actions. Escape or Close Switcher in the options menu dismisses the chooser. Real MacTools windows are included; confirmed windowless apps and transient helper surfaces are excluded.
- Previews support pinch zoom up to 4×, panning, double-click to fit, and menu/keyboard alternatives. The chooser acquires focus for physical pinch in both cycling and search modes. Cancelling restores the original application only while the chooser owns focus.
- Native glass follows system appearance and accessibility preferences, with a material fallback. Search matches use contrasting foreground/background colors. Settings explain Behavior and Order choices through full-row selection.

## Cross-Space activation and previews

Application activation alone can bring forward the wrong window, while a successful `AXWindows` read can omit a real window on another Space. The plugin validates WindowServer ownership and exact window identity before resolving the missing AX window root. The selected-window lookup is bounded and runs on the owning process worker; descendants reporting a containing window ID are rejected.

The optional native activation bridge fronts the selected window ID, sends a balanced window-addressed focus event, then applies AX focus/raise. Minimized windows restore once before fronting. Completion checks the exact focused window, onscreen visibility, and current-Space membership across displays when available. Visibility alone can become true before the Space transition completes. Cancellation guards each stage; verification does not repeatedly issue focus requests. No windows are created, resized, or moved between Spaces.

Off-Space previews use an optional exact-ID WindowServer capture fallback when ScreenCaptureKit cannot supply pixels. Capture remains permission-gated, owner-validated, memory-only, and bounded by the existing image-size and request-lifecycle limits. Missing private activation symbols preserve public application activation; unavailable capture capabilities yield an unavailable preview.

Discovery distinguishes a missing Core Graphics window-name field from an explicitly empty title. Without Screen Recording permission, eligible off-Space candidates with positive Space membership remain selectable and display their app name. Explicitly empty or malformed titles retain the existing AX-confirmation requirement. Display fallback does not become identity metadata; activation still checks exact ownership and AX window identity, and previews still require Screen Recording permission.

## Reliability and compatibility

Per-process Accessibility workers bound discovery work and preserve incomplete-scan uncertainty. Activation observes intervening foreground application changes from the moment of selection, before chooser dismissal or asynchronous validation; superseded work stops submitting actions. Already submitted system requests cannot be recalled, and application activation notifications do not identify every same-app window change. Destructive menu actions retain their original target and require live identity validation; remote AX lookup is used only for activation.

Preview requests remain serialized. Uncached selections wait for an 80 ms quiet period; cached images and zoom-detail requests do not. Old pixels blur immediately and remain until replacement pixels arrive, then swap without a fade or minimum hold time. Empty selections, errors, permission loss, and dismissal clear them; Reduce Motion skips the blurred hold. Gesture recognition remains enabled throughout screenshot replacement. Timeouts cannot overwrite a newer selection or create parallel captures. Standard captures are capped at a 1,600-pixel edge, detail captures at 3,200 pixels.

Requires **MacTools 1.3.1 / PluginKit v6**. Host shortcut handling and plugin changes should ship together; no new public PluginKit API is introduced. Generated plugin/action data is verified by local script tests and CI. See the [development notes](docs/plugins/window-switcher.md) and [interaction contract](docs/superpowers/specs/2026-09-10-window-switcher-redesign.md). Related: #414, #415, #416.

## Validation

- CI layout-test repair `c4367d98`: **56 Window Switcher session tests passed locally**. Automatic sizing is checked against the actual display; the two-row geometry test now sets its intended size explicitly. New native collection coverage verifies four/five/six columns, both scrollbar styles, overflow reachability, and stable selection. This fixes hosted assertions that assumed the developer display's column count; production code is unchanged.

- Latest title-permission repair: **109 focused XCTest tests passed**, including cold discovery without name metadata, title availability changes, stable selection, ghost-surface rejection, and denied/revoked preview permission. Two earlier regression methods were moved out of a helper class into the XCTest class so they actually run; active-Space membership and balanced focus-event tests now pass as discovered tests. Whitespace and changelog checks passed.
- Latest PR branch merged with `main` at `d60c5819`: **`make ci` passed**, with **5,160 XCTest passes, 4 skips, 0 failures**, **268 script tests**, generated-data/changelog validation, and PluginKit v6 binary compatibility. The Makefile conflict preserves main's serial execution and test timeouts, keeps native drag fixtures opt-in as established upstream, and retains this PR's generated-data check and explicit compatibility-test build path. Window Switcher source is unchanged by this merge.
- Earlier combined-integration validation had a clipboard helper EOF that passed on retry and a Dashboard LTR synthetic-drag fixture failure. The latest standalone CI pass above supersedes that snapshot as the current automated gate; it does not establish native drag acceptance. The opt-in drag fixture was not rerun for this build/CI-only reconciliation.
- On macOS 27, the earlier relinked production-code diagnostic selected cold-discovered Chrome windows absent from the initial AX list. The regular desktop and fullscreen destinations passed exact focus and active-Space verification in approximately 0.5 seconds; both previews loaded in approximately 0.1 seconds. Mouse-button state was released afterward. These observations do not validate the newly repaired Screen Recording-denied scenario and are not latency guarantees.
- The latest combined Debug app was clean-built, installed, and relaunched. Host/plugin hashes matched the build, signature verification passed, and the running app loaded the updated plugin. The user previously confirmed the cross-Space fix and physical pinch behavior. The latest title-permission repair still needs native acceptance with Screen Recording denied; existing permission settings were not changed.
- Hosted checks for `c4367d98` **passed**: [Build run 34963785441](https://github.com/ggbond268/MacTools/actions/runs/34963785441), including website verification, script tests, XCTest, PluginKit compatibility, and unsigned Nightly configuration verification. The prior hosted failure was limited to the two display-dependent layout assertions repaired above. macOS 14/26 rendering, physical IME/VoiceOver, multiple displays, Split View, fullscreen-to-desktop, native minimized cross-Space restoration, Stage Manager, and interrupted transitions are not exhaustively covered by local acceptance checks.

## Reviewer focus

1. Exact identity and lifetime validation when AX lists omit a window; balanced native focus events and cancellation boundaries.
2. Optional private API fallbacks and destination verification across separate display Spaces.
3. Direct Keys migration, explicit search transitions, and host shortcut compatibility.
4. Preview permission/generation handling, bounded capture, and continuous gesture recognition.


ggbond268:
Thanks for the work on this! I found one regression to address before merging:

**Scenario:** Accessibility is granted, Screen Recording is not, and a window on another Space or in fullscreen has not yet been exposed through `AXWindows`.

**Cause:** macOS withholds `kCGWindowName` without Screen Recording permission ([Apple's explanation](https://developer.apple.com/videos/play/wwdc2019/701/)). The new [filter in `mergeAllSpacesEntries`](https://github.com/ggbond268/MacTools/blob/0c17e878cbedbf571edd18608420443ee69fe85d/Plugins/WindowSwitcher/Sources/WindowSwitcherAppCatalog.swift#L756-L760) drops any record with no title unless AX previously confirmed it, even when the window has positive Space membership. This makes valid windows disappear from the chooser. A focused regression test confirms that the previous implementation retains the same record while the new implementation drops it.

**Suggested fix:** Distinguish unavailable title metadata from an actually empty title. Keep otherwise valid other-Space candidates using the existing window identity and Space checks, with the app name as a display fallback. Preserve exact AX identity/role validation before activation and filtering of invalid compositor surfaces. Please add coverage for this permission combination alongside the existing surface-filtering tests.

Screen Recording should remain optional for previews; basic window switching should work with Accessibility alone.

PR #425, opened 2026-09-12
feat(plugin): add AI Assistant plugin with floating panel and text capture pipeline

lyhu:
### Summary
Adds the new **AI Assistant** (`AIAssistant`) dynamic plugin to MacTools:
- **Global Text Capture Pipeline**: Captures selected text across macOS applications via Accessibility API, simulated copy fallback, and AppleScript for supported browsers.
- **Floating Processing Panel**: Clean floating panel with instant auto-save prompt templates, inline editable source text (supports double-click to edit and reprocess), and smart result view with 10-line auto-sizing and copy feedback.
- **Provider & Template Settings**: Supports OpenAI-compatible endpoints, custom models, thinking/reasoning parameters, and per-template sampling temperatures.
- **Inline Shortcut Management**: Embedded shortcut recorder within each template card, equipped with system hotkey protection (warning on ⌘C, ⌘V, etc.) and duplicate shortcut collision detection.

### User Impact & Documentation
- Added changelog fragment under `changes/unreleased/ai-assistant.md`.
- Updated feature list in `README.md` and `README.zh-CN.md`.
- Fully localized in English and Simplified Chinese (`Localizable.xcstrings`).

### Verification
- Added comprehensive unit test suites under `Plugins/AIAssistant/Tests/` covering coordinator, client, provider profile, secret store, prompt store, and plugin runtime behaviors.
- Verified compilation and test pass via:
  ```bash
  xcodebuild -project MacTools.xcodeproj -scheme MacTools -configuration Debug -derivedDataPath build/DerivedData test -only-testing:MacToolsTests/AIAssistantPluginTests -quiet


ggbond268:
Thanks for working on this! I found a few issues that should be addressed before merging:

1. **Preserve existing default shortcuts.** The shared resolver now passes `default: nil`, so Physical Clean Mode loses its default exit binding and refuses to start unless the user configures one. Please restore the resolver's existing behavior and avoid calling it from the AI plugin's `shortcutDefinitions` getter. If the legacy IDs never shipped, remove that migration; otherwise, migrate once while preserving custom and disabled bindings.

2. **Fix the new-template naming loop.** The localized name contains `%d`, but `localization.string` does not format the index. Keeping the first new template's default name and adding another causes an infinite loop on the main thread. Please use `localization.format` and test consecutive additions with the plugin's actual localization resources.

3. **Keep user content out of logs.** Captured text is logged with `.public`, and raw provider error messages are also logged publicly. Please remove these payloads and retain only diagnostic metadata such as capture strategy, character count, and HTTP status.

4. **Protect the simulated-copy operation.** Cancelled captures continue running, and clipboard restoration can overwrite content copied in the meantime. Please serialize copy operations, honor cancellation, use the host's captured application target, and send the copy event only once. Avoid restoring over subsequent clipboard changes. A changed clipboard alone does not establish who wrote it, so ambiguous input should not be sent automatically.

5. **Bound blocking capture work.** AX calls and the System Events AppleScript fallback run synchronously on the main thread. Please move blocking work off the UI thread and apply execution timeouts. Script execution should be serialized; cancelling a Swift task alone does not stop an already-running synchronous script.

Focused tests reproduced the shortcut, cancellation, and clipboard issues, and a separate probe reproduced the template loop. Two existing HTTP 401 tests also fail. Please also resolve the plugin metadata/minimum-host checks and the changelog validation failure.

Could you take a look and update the PR? Happy to review the changes.

xcv58:
Following up on item 1 in the earlier review: the dependency between AI Assistant's shortcut definitions and the shared resolver is the part to address in this PR.

`AIAssistantPlugin.shortcutDefinitions` calls `defaultBinding` → `migratedDefaultBinding` → the host's shortcut-binding resolver. The existing host resolver enumerates shortcut definitions, so calling it from that getter creates a recursive dependency.

This PR avoids that recursion by reading storage directly with `default: nil`, but that changes the shared resolver's behavior for other plugins, including Physical Clean Mode's default exit shortcut.

Please keep the existing default/custom/disabled binding semantics and move any necessary legacy migration outside `shortcutDefinitions`. If those legacy IDs never shipped, removing the migration would be simpler.

Regression coverage should verify that reading AI Assistant's shortcut definitions does not re-enter the getter, existing custom and disabled assignments survive migration, and other plugins still receive their default bindings.

xcv58 (commented):
Thanks and I have tried this plugin locally! I think a few UX improvements would make this much more dependable:

1. **Make dismissal non-destructive.** Losing focus currently closes the panel and cancels the task, with no way to reopen the result. Separate **Hide**, **Stop**, and **Discard**, preserve the current session, and add a **Reopen last session/result** action that does not recapture text or make another API request.

2. **Fix error recovery.** After capture fails, **Retry** does nothing because `lastSourceText` is nil. Retry should attempt capture again, with a visible **Paste/type text instead** option. Show specific configuration errors rather than replacing them with a generic message.

3. **Preserve the previous result during reruns.** Editing the source currently clears the successful output before the replacement succeeds. Keep the previous output recoverable, especially if the new request fails. Rename **Confirm** to **Apply and rerun**.

4. **Keep window placement stable.** Updating processing/results should not reapply an old/default frame or take keyboard focus again. Dragging the panel during generation should survive completion.

5. **Make template saving predictable.** Text edits currently update view state, while Done and other controls trigger persistence. Use explicit **Save/Cancel** or consistent autosave with a visible save status.

Full persistent history, pinning, and resizing can be follow-ups; basic session recovery should come first. Please add regression coverage for these interaction paths.

Thank you again!

lyhu:
Thanks both for the detailed reviews. All feedback is addressed in 2fb69ad4 (single commit on top of 20588dfa).

**Correctness**
1. *Shortcut definitions / resolver recursion* — `PluginHost.legacyResolvedBinding(for:)` resolves through the enumerated descriptor and `descriptor.definition.defaultBinding` again, so Physical Clean Mode and other plugins keep their defaults, and custom/disabled assignments survive. `AIAssistantPlugin.shortcutDefinitions` no longer touches the resolver: it maps prompt IDs straight to `AIAssistantConstants.Defaults`. The `process-translate` / `process-summary` / `process-text` migration is removed — those IDs only ever existed on this unreleased branch, never in `main`. Regression: `AIAssistantPluginTests.testShortcutDefinitionsIgnoreHostLegacyResolverAndKeepDefaultBindings` proves the resolver hook is never called, including after saving a custom prompt.
2. *New-template naming loop* — `AIAssistantPromptStore.makeNewPrompt` now formats the localized pattern via `localization.format` and keeps a defensive bound; `AIAssistantPromptStoreTests` covers consecutive additions against the real resource values.
3. *Logs* — capture and provider logs keep capture strategy, character count, and HTTP status only; provider error bodies are no longer logged, and 401 logs a bare `.unauthorized`.
4. *Simulated copy* — `CaptureSerializationGate` serializes captures, a capture cancelled while queued never sends a copy event, the sender runs once with the host's captured `context.frontmostApplicationProcessIdentifier` and picks exactly one destination, and `restoreIfUnchanged` restores only while the pasteboard change count still matches, so text copied afterwards is never clobbered. 14 tests in `AIAssistantSimulatedCopyCaptureTests`.
5. *Blocking work* — AX probing runs in a detached task with a 1s watchdog and a single-resume continuation box; the System Events fallback goes through `SerializedAppleScriptRunner` (off the main thread, serialized, watchdog with recovery).

**UX**
1. Dismissal is non-destructive: Esc / xmark / focus loss = Hide (session kept), Stop cancels and restores the last finished result, trash = Discard, and pressing the same prompt shortcut on a hidden session reopens it without recapturing or issuing another request.
2. Retry recaptures when `lastSourceText` is nil, "Type Text Instead" allows manual entry, and configuration errors surface verbatim instead of one generic message.
3. The previous result stays visible and recoverable during a rerun; Confirm is renamed to Apply and Rerun.
4. Processing and result updates no longer reapply a stored frame or take focus, so dragging the panel during generation survives completion.
5. Template editing uses explicit Save/Cancel with a draft snapshot, so Cancel reverts without persisting.

Regression coverage for these paths lives in `AIAssistantCoordinatorTests` (hide + reopen without recapture, stop restoring the finished result, discard/close, retry recapture, retained result during rerun).

**Metadata / changelog** — `pluginKitVersion 7`, `minHostVersion "1.3.1"`, full marketplace metadata; the changelog fragment is 217 characters / 2 sentences and passes `make validate-changelog`. The localization catalog is now complete as well: 135 keys, no missing or unused entries, en / zh-Hans / zh-Hant all translated.

Verification: `make build` exit 0; focused plugin suite 59 passed / 0 failed (both HTTP 401 tests included); `make script-tests` 271 tests OK.

ggbond268:
Thanks for the update! I rechecked the latest changes and found a few remaining issues:

1. **Clipboard fallback can capture unrelated text.**  
   If the simulated copy fails and another app updates the clipboard, that content can be treated as the selection, sent to AI, and overwritten when the old clipboard is restored. Please preserve the host-captured target, avoid untargeted copy retries, and ask for confirmation before sending clipboard fallback text. A change count alone cannot establish who wrote the content.

2. **Timeouts leave the underlying work running.**  
   AX calls continue after task cancellation, and AppleScript execution continues after its timeout. Creating a replacement queue allows more unfinished work to accumulate. I confirmed that a script can still execute after the caller receives a timeout. Please use bounded background execution with native AX timeouts and cancellable script execution, rather than replacing queues when they become blocked.

3. **The HTTP host check accepts some public domains as private addresses.**  
   For example, `http://10.0.0.1.example.com` passes validation because nonnumeric parts are discarded. Please validate the complete IPv4 address before checking private ranges. This is a small validation fix; it does not establish that credentials are currently leaking.

4. **A hidden panel reopens when the request finishes.**  
   Pressing Esc during processing hides the panel, but completion shows it again. Please update the retained result without reopening the panel until the user requests it. Reusing the shared panel presentation and dismissal components would also help align this plugin with the current conventions.

These should be manageable within the plugin, with clipboard handling and cancellation being the main priorities.

lyhu:
@ggbond268 Thanks for the follow-up. I pushed `29ad0d47` and updated the PR description. The four remaining points are addressed:

1. **Clipboard:** simulated copy now targets only the application PID captured by the host and sends no untargeted System Events retry. Clipboard-derived text requires confirmation before an AI request. We no longer restore the old clipboard snapshot: when the first change came from another app, its ownership cannot be established, so restoring could overwrite that app's copy. The latest clipboard content remains in place.
2. **Underlying work:** target-app AX elements have native messaging timeouts, and the unbounded system-wide AX query was removed. Browser AppleScript runs in a serialized child process. Cancellation and timeout terminate that process before the next run is allowed; stdout is read concurrently so large output cannot block process exit.
3. **HTTP host validation:** the full dotted-quad address is parsed before private-range checks; `10.0.0.1.example.com` is rejected.
4. **Hidden panel:** completion updates the retained snapshot without showing the panel. It reopens only on an explicit user action. Retry also cannot bypass clipboard confirmation.

Verification: coordinator tests 15/15, simulated-copy tests 14/14, script tests 221/221, and changelog validation passed. The shared-panel component refactor is not part of this focused fix; the hide/reopen behavior has regression coverage.

lyhu:
@ggbond268 Follow-up after testing the installed Dev app in Chrome: the "Possible Clipboard Content" prompt still appeared for the copy-first workflow. This is expected in selection mode because its simulated-copy fallback cannot prove who wrote the clipboard; it was not caused by an old app copy.

I pushed `7f2d9fde` with an explicit **Use Clipboard** setting for that workflow. When enabled, copying text and pressing a prompt shortcut reads the current clipboard directly and starts processing without the ambiguous-source prompt. Selection mode remains the default and still requires confirmation for simulated-copy results, so unrelated clipboard changes are not silently sent to AI. An empty clipboard shows an error, and Retry reads it again.

The coordinator and plugin tests passed (33 tests), followed by 17 plugin tests after the hidden-session adjustment. Changelog validation passed. I installed and launched the new Debug build from the sole `~/Applications/MacTools Dev.app` path for manual retesting.

ggbond268:
Thanks for the follow-up fixes! The HTTP validation, hidden-panel behavior, and script timeout handling look improved. Two points still need attention:

1. **The latest commit removes the clipboard confirmation safeguard.**  
   In selection mode, another app can update the clipboard while simulated copy is waiting. That unrelated text can then be sent to AI automatically.  
   Please restore confirmation for simulated-copy results in selection mode. The explicit **Use Clipboard** setting can continue sending directly, since the user has deliberately chosen that input source.

2. **The capture target still comes from the current foreground app.**  
   The coordinator reads `NSWorkspace.shared.frontmostApplication` inside the asynchronous task rather than using the host’s saved target. When a MacTools surface takes focus, this can lose the original application and cause capture to fail.  
   Please adopt the existing `PluginFocusedWindowTargetConsuming` protocol, resolve the target when the action starts, and keep that target throughout the capture operation.

Both changes should stay within the plugin and reuse the existing host APIs.

lyhu:
> Thanks for the follow-up fixes! The HTTP validation, hidden-panel behavior, and script timeout handling look improved. Two points still need attention:
> 
> 1. **The latest commit removes the clipboard confirmation safeguard.**
>    In selection mode, another app can update the clipboard while simulated copy is waiting. That unrelated text can then be sent to AI automatically.
>    Please restore confirmation for simulated-copy results in selection mode. The explicit **Use Clipboard** setting can continue sending directly, since the user has deliberately chosen that input source.
> 2. **The capture target still comes from the current foreground app.**
>    The coordinator reads `NSWorkspace.shared.frontmostApplication` inside the asynchronous task rather than using the host’s saved target. When a MacTools surface takes focus, this can lose the original application and cause capture to fail.
>    Please adopt the existing `PluginFocusedWindowTargetConsuming` protocol, resolve the target when the action starts, and keep that target throughout the capture operation.
> 
> Both changes should stay within the plugin and reuse the existing host APIs.



Thanks for the review! Addressed both points in `bb089235`.
Simulated-copy results in selection mode now require confirmation before any text is sent to AI. The plugin also uses `PluginFocusedWindowTargetConsuming` to resolve the host-saved target when the action starts and keeps it throughout capture.
The same shortcut now prefers directly readable selected text. When Use Clipboard is enabled, it falls back to the clipboard text captured at shortcut time only if no selection was captured and the clipboard remained unchanged. Simulated-copy results still require confirmation.
Verified: 21 coordinator tests, 19 plugin tests, and 221 script tests passed.

ggbond268:
Thanks for the follow-up fixes. I re-reviewed `bb089235`; the earlier capture-target, confirmation, and timeout changes look improved. I found the following remaining issues and recommend addressing all of them:

1. **Release inactive global shortcuts.** AI Assistant keeps its global shortcut definitions when the main switch is paused or an individual prompt is disabled. The handler ignores the event, but the host still registers the binding. With the default ⌥1/⌥2/⌥3 bindings, this can consume keys in other apps and conflict with other plugins. Please keep saved bindings editable while unregistering inactive shortcuts, then restore them when re-enabled. A focused host registration test for both switches would help.

2. **Make prompt Save/Cancel transactional.** Field bindings edit the shared `prompts` array, and adding, moving, or toggling another prompt persists that whole array. As a result, cancelling an edit can leave its draft saved; adding a new prompt also saves it before its own Save, so Cancel does not remove it. Please keep each edit in a separate draft and persist it only after a successful Save. Other prompt actions should not commit that draft.

3. **Preserve an intentionally empty prompt list.** Deleting the final prompt saves `[]`, but `loadPrompts()` treats an empty decoded array like missing storage and recreates the defaults. Please distinguish “no saved value” from “saved empty list,” or prevent deleting the last prompt if an empty list is unsupported.

4. **Keep the editor open when Save fails.** `saveEditing()` closes the editor even when `onPromptsChanged` rejects an enabled template without `{{text}}`. Please close and discard the draft only after persistence succeeds, leaving the validation message and editable fields visible on failure.

5. **Handle simulated-copy history explicitly.** I understand why the fallback leaves the latest clipboard content in place. However, when Clipboard History is enabled, the synthetic ⌘C can also be stored as a history item before the user confirms sending anything to AI. Please address that cross-plugin privacy effect, for example with scoped coordination that suppresses this internal capture, or an explicit in-app opt-in for simulated copy. Blindly restoring the previous clipboard would reintroduce the overwrite risk.

6. **Use English for code comments.** Several new Swift comments are in Chinese, which conflicts with the repository's `AGENTS.md` convention. Please translate those comments while updating the files.

Focused regression coverage for the shortcut and prompt-state paths would make the intended behavior clear. Thanks again for iterating on this.

PR #434, opened 2026-09-15
feat: add Duo Status plugin with optional primary icon replacement

ryanbuu:
## Summary

- Package Duo Status as an independent, optional plugin under `Plugins/DuoStatus`, adapting the original Duo artwork and monitoring by @ryanbuu.
- Replace the prototype visibility switch with a declarative segmented setting: **Separate Icon** (default) or **Replace App Icon**. No host-side source picker, Feature Panel entry, or Dashboard component is added.
- Add optional PluginKit v6 icon contracts and a Core coordinator for exclusive primary-icon ownership. Competing requests return a typed error naming the owner and keep the current mode unchanged; plugins cannot force takeover or impersonate another provider.
- Keep the host's existing status item, saved position, panel anchor, click routing, and automation badge. Uninstall, disable, or isolation restores the original configured icon without overwriting custom images or animations. Startup and update reservations preserve the selection, while revoked contexts and stale callbacks cannot affect replacement instances.
- Share the plugin monitor across both modes, read cached snapshots in render calls, coalesce primary-icon notifications over 150 ms, and skip unchanged frames. Icon ticks do not rebuild the host's general plugin state. Pause sampling while inactive or asleep.
- Include copy in all 11 supported locales, minimum-host compatibility inventory, adjacent regression tests, plugin packaging, documentation, website catalog data, and separate app/plugin changelog entries. Adopting plugins require host 1.3.1 or later; the existing PluginKit v6 binary contract is preserved.

With a separate icon, hover for status details and click either mouse button to open plugin settings. Replacement keeps the normal MacTools panel actions. Network status describes the local connection, not verified internet reachability. No SSID collection, location permission, or external connectivity probe is introduced.

## Validation

- `make generate`
- Final focused XCTest: **92 passed**, covering Duo settings/lifecycle, artwork, monitoring, hardware normalization, icon arbitration, host click routing, original-icon settings, and dynamic plugin teardown
- `make script-tests`: **269 passed**, including the new public-API minimum-host inventory
- PluginKit v6 binary compatibility client: passed
- Unsigned Debug build and local Duo Status `.mactoolsplugin` packaging: passed
- Website generated-data check, build, and tests: **4 passed**
- `make validate-changelog` and `git diff --check`: passed

An earlier full-suite run was not green: `AppleShortcutsCommandRunnerTests.testTimeoutForceKillsTermIgnoringDescendantAndDrainsPipes` failed because its `child.pid` fixture was not generated; one command-palette input test was terminated but passed on rerun. Those modules were left unchanged. Subsequent validation is limited to relevant non-UI tests; no further full-suite or UI test runs are included in this update.

## Original visual reference

The initial host-icon proposal is retained below as the visual reference. This revision defaults to an independent plugin icon and optionally uses the shared host interface to replace the primary icon.

<details>
<summary>Original proposal screenshot</summary>

Original Duo host-icon proposal, retained as a visual reference

</details>


xcv58 (approved):
lgtm, I love this.

PR #437, opened 2026-09-16
feat(DeviceBattery): JBL Sense Lite split battery via ExcelPoint BLE

AndrewDi:
Add separate left-earbud, right-earbud, and charging-case battery levels for JBL Sense Lite using ExcelPoint BLE notifications.

- Bind each peripheral to one unambiguous connected device, supporting the JBL `-LE` suffix and retaining existing device classification and system battery fallbacks.
- Coalesce connection events without interrupting an active read. Refresh the connection snapshot and scan only newly connected devices, preventing reader-generated events from causing repeated scans.
- Preserve periodic refreshes, cached readings, conditional discovery, and the existing three-second scan timeout. Limit ExcelPoint discovery to JBL candidates.

Validation: 147 focused Device Battery tests and five local regression probes passed, including scan-feedback, system-fallback, and parser checks. Changelog validation passed. The latest discovery and lifecycle changes have not been retested on physical JBL hardware.

Contributor screenshot:

JBL Sense Lite earbud and case battery levels


AndrewDi:
@ggbond268 @ggbond2077 PTAL

ggbond268:
After a closer look at how this integrates with the existing battery collection paths, I found three issues worth addressing before merging. The separate protocol parser makes sense; the main concerns are in the shared scanning and fallback behavior.

1. **Preserve existing battery readings and device identity.** The new filters remove all devices whose names start with `JBL` from the system battery results and target list. This drops working readings for unsupported models or when the ExcelPoint read fails. Please retain those readings and targets, associate ExcelPoint results with the same device identity, and let the existing deduplication logic prefer left/right readings when available.

2. **Limit connections to known, connected devices.** The scanner now accepts any nearby peripheral whose name starts with `JBL`, even without a matching target. Its readings can then appear in the device list. Please add JBL candidates to the existing scan plan, preserve the target-matching checks, and verify the ExcelPoint service before using its data.

3. **Keep scanning conditional and bounded.** Every active refresh now starts an unfiltered scan, even with no eligible devices, and an empty scan waits five seconds. The previous timeout was three seconds. Please restore the existing scan policy: return immediately without targets, query connected peripherals using explicit service UUIDs, scan only when needed, and stop once the targets are resolved. If JBL needs a longer timeout, please scope that change to its collection path.

A focused regression test reproduces the first issue: the same JBL profile retains its battery reading on `main` but loses it in this PR. Please add coverage for fallback readings, deduplication, unmatched devices, and empty scan plans, while preserving the existing AirPods and standard GATT paths.

These changes should fit within the existing scanner and normalization code without a broader refactor. Happy to review the update.

AndrewDi:
## JBL Sense Lite BLE Battery Read Implementation

### Overview
Added split battery support for JBL Sense Lite earbuds (left, right, charging case) to the DeviceBattery plugin via the proprietary ExcelPoint BLE GATT service.

### Implementation Details
- **ExcelPoint protocol parser** (`JBLSenseLiteBLEBatteryParser.swift`): Parses 55-byte notification packets to extract left/right/case battery levels
- **Scanner integration** (`DeviceBatterySampler.swift`):
  - Registers ExcelPoint service UUID `65786365-6C70-6F69-6E74-2E636F6D0000`
  - Supports both pre-connected peripherals and scan-based discovery
  - Uses prefix matching to associate BLE peripherals with system_profiler devices
- **UI integration** (`DeviceBatteryComponentView.swift`): `headphones` icon for earbuds, `bolt.circle` icon for charging case
- **Localization** (`Localizable.xcstrings`): Added component translations for left ear, right ear, and charging case
- **Regression tests** (`DeviceBatterySamplingPolicyTests.swift`): 8 tests covering parsing, scan policy, deduplication, and fallback behavior

### Key Fixes
1. **ViewModel scan cancellation**: JBL BLE connections triggered onConnectionChange, which cancelled the running scanner Task and discarded results. Fixed by guarding handleBluetoothConnectionChange with a bluetoothTask == nil check.
2. **Device identity mismatch**: ExcelPoint component items used a different deviceIdentity than the system_profiler aggregate, preventing deduplication. Fixed by adding jblTargetMatching fallback in gattBatteryItems.
3. **Slow scan completion**: target(for:) only used uniqueTarget for exact name matching, so the JBL peripheral ("JBL Sense Lite-LE") never matched the target ("JBL Sense Lite"). The scanner had to wait for the full 5-second timeout. Fixed by adding jblTargetMatching fallback in target(for:), allowing the scanner to finish immediately after receiving data.

### Test Results
- Parser validation: left=100%, right=99%, case=49% ✓
- Scan policy: JBL device correctly identified as GATT target ✓
- Deduplication: Component items preferred over aggregate item ✓
- UI display: Split battery levels displayed correctly, no duplicates ✓

AndrewDi:
@ggbond268 fix all issue,PTAL again

ggbond268:
Thanks for the update. The `-LE` completion fix looks correct. There is still a blocking issue in the connection-change handler:

`bluetoothTask` represents the long-running polling loop, including the time between collections. Therefore, `guard bluetoothTask == nil` ignores normal connection and disconnection events while monitoring is active. The existing event-refresh test still reproduces this regression.

Please handle this through the existing collection lifecycle:

- Use `isCollecting(.bluetooth)` to identify an active collection. If one is running, record a pending connection refresh and process it afterward. Otherwise, use the existing debounce behavior. Multiple events should coalesce into one pending refresh.
- For connection events, refresh the actual connection snapshot first. If the connected device set is unchanged, retain valid readings without starting another active scan. Update disconnected devices and actively read newly connected eligible targets. Periodic and manual refreshes can retain their existing behavior.
- Keep connection-snapshot refresh separate from clearing battery data. `forceProfileRefresh` currently clears the entire supplemental cache, which would discard freshly collected JBL readings. Preserve valid readings for devices that remain connected and apply the existing expiration rules.

This avoids both cancelling an in-progress read and creating a cycle where scanning triggers an event that immediately starts another scan.

Please also restore the original conditional scanning policy and three-second timeout unless device testing demonstrates a JBL-specific need. Resolve each peripheral to its target once, then reuse that binding for result generation and completion tracking.

This should fit within the current structure: the ViewModel schedules refreshes, the sampler determines which devices need reading, and the scanner performs bounded I/O.

AndrewDi:
Thanks for the detailed review. All three issues are addressed:

### 1. Connection-change lifecycle (`handleBluetoothConnectionChange`)

Replaced `guard bluetoothTask == nil` with `isCollecting(.bluetooth)`. The previous guard treated the entire polling loop as "in progress", ignoring normal connection events while monitoring was active. Now:
- If a bluetooth collection is in progress, a `pendingConnectionRefresh` flag is set and coalesced into the next loop iteration.
- The loop picks it up after the current collection finishes, avoiding cancellation of in-progress scans.
- `pendingConnectionRefresh` is reset in `cancelSampling`.

### 2. Scan policy and timeout restored

- **Scan guard**: Removed the `|| !peripheralHasJBLService.isEmpty` escape hatch. The scanner now returns immediately when `expectedTargetIDs` is empty, matching the original policy. JBL peripherals are already retrieved via `retrieveConnectedPeripherals(withServices: [Self.jblExcelPointService])` before the guard.
- **Timeout**: Restored to 3 seconds (was 5s).

### 3. Supplemental cache preserved on profile refresh

Removed `supplementalItemCache.removeAll()` from `collectBluetoothDevices` when `forceProfileRefresh` is true. The cache's `update` method already handles expiration via `retentionInterval` and drops entries for disconnected targets. Clearing it entirely was discarding freshly collected JBL readings when the connection snapshot refreshed.

All existing tests pass (DeviceBatterySamplingPolicyTests + DeviceBatteryViewModelTests including `testSystemEventsRefreshOnlyTheirSourceAndDebounceBluetooth`).

AndrewDi:
@ggbond268 PTAL again

PR #441, opened 2026-09-18
Clipboard: add per-item paste shortcuts

### LINKED ISSUE #351: Clipboard History: add a temporary Quick Paste slot
## Problem

Some short-lived workflows require repeatedly pasting one specific Clipboard History or Saved Library item while the user continues copying changing values.

For example, a user may assign an email snippet containing `{{clipboard}}`, copy a recipient name, paste the expanded template through one shortcut, copy another name, and repeat. Permanent snippet shortcuts are heavier than needed, while reopening Clipboard History for every paste interrupts the workflow.

## Proposed experience

Add a **Temporary Quick Paste** slot to Clipboard History:

- Configure the slot's global shortcut once in Clipboard History settings.
- From a History, Saved, or Snippet item, choose **Assign to Quick Paste**.
- Choose an expiration of 5, 10, or 30 minutes.
- Pressing the shortcut repeatedly pastes the assigned item without consuming the assignment.
- Snippet variables, including `{{clipboard}}`, are evaluated at each paste so newly copied values are used.
- Ordinary clipboard changes do not reset the slot.
- Assigning another item replaces the current assignment after clear confirmation.
- A lightweight HUD confirms assignment, replacement, remaining time, manual clearing, and expiration.
- Expiration, manual clearing, plugin deactivation, or app restart clears the assigned content while retaining the configured shortcut.

## MVP scope

- One Quick Paste slot.
- History, Saved clips, and Snippets can be assigned.
- Expiration choices: 5, 10, and 30 minutes.
- Manual **Clear Quick Paste** action.
- The existing paste-target safety and Accessibility guidance apply.
- Shortcut conflicts use the existing MacTools shortcut validation.
- Assignment state is shown in Clipboard History settings and the item action palette.

## Non-goals for the first version

- Arbitrary new global shortcut registration for every item.
- Multiple concurrent slots.
- Permanent assignments.
- Synchronizing the assignment across Macs.
- Clearing the slot when unrelated clipboard content changes.

## Acceptance criteria

- Repeated shortcut presses paste the same assigned literal item until it expires or is cleared.
- A Snippet expands against the current clipboard on every invocation.
- Copying a new value does not cancel the assignment.
- Expired or cleared assignments cannot paste stale content.
- Restarting MacTools clears the temporary assignment.
- The HUD and settings identify what is assigned and when it expires without exposing sensitive content unnecessarily.
- Assignment, replacement, expiration, and clearing have regression tests.

xcv58:
## Summary

- Let each Clipboard History item and Saved clip have its own global paste shortcut. Items with one plain-text representation and Snippets show one Paste Text shortcut. When paste results can differ, including multi-item clipboard entries, the sheet explains Paste Original and Paste as Plain Text and offers separate keys, timers, and removal controls. Existing second shortcuts on text-only items remain visible and removable.
- Offer 5 minutes, 1 hour, 1 day, and Until removed for every item. Timed shortcuts protect History items from automatic cleanup through their duration, including across app restarts. Choosing Until removed saves an unsaved History item.
- Paste the item's latest content, preserving rich content for Original and using the existing visible-text conversion for Plain Text. Expand Snippet variables on each use. Remove assignments when items disappear or timers expire; preserve them through cancelled or temporary payload-read failures.
- Reject shortcuts already owned by another plugin or action, naming the owner. The item sheet links to Actions & Shortcuts for resolution. If a later plugin default collides, an explicit item assignment retains the key and the other shortcut shows a conflict.
- Harden shortcut lifecycle handling after follow-up review: rejected timed-to-lasting updates preserve the original assignment and retained item, transient initial load failures do not prune persisted shortcuts, rapid item shortcuts wait for each paste to be consumed, and shortcut definitions are cached instead of repeatedly scanning large histories.
- Replace the original single Quick Paste slot with the per-item design agreed after issue #351 was filed.

Closes #351.

## Validation

- Before this follow-up, 210 affected Clipboard backup, plugin, controller, and shortcut-store tests passed, and hosted CI passed on that exact head, including native tests, PluginKit compatibility, and unsigned Nightly validation. On `d85d07c1`, the Clipboard History test class passed with one intentional visual-capture skip, all five shortcut-store tests passed, focused regression tests passed, and diff checks passed. Hosted CI for the new head is pending.
- The first independent review found two issues: multi-item plain-text entries had lost the second paste option, and Plain Text file-path shortcuts unnecessarily required the original files to exist. Both were fixed and covered by regression tests; follow-up review found no further issue at that revision.
- A second independent review found two concurrency issues: overlapping Original and Plain Text saves could reject one assignment, and an in-flight paste could continue after plugin deactivation. Both were fixed and covered by regression tests. Its follow-ups found backup-suspension races that could finish an assignment or paste across pause and resume, or leave an automatically Saved item after a rejected assignment. Those are fixed with three more regression tests. The reviewer rechecked the exact head and found no actionable issues.
- A third independent review found that a provisional Saved clip could enter a backup before a rejected shortcut rolls it back. The backup now omits that exact provisional Saved metadata while preserving History and any separately saved content. Its follow-up found the same provisional metadata could enter restore staging or the local rollback snapshot; both paths now omit it. Archive, restore, rollback, and suspension-window regression tests passed. The reviewer rechecked that revision and found no actionable issues.
- The latest review comment identified four additional cases. Regression tests now cover rejected lasting updates, temporary initial-load failures, rapid A/B item shortcuts, and bounded shortcut-definition lookup with 10,000 History items and 100 shortcuts. The performance test fails on the pre-fix head at 37.952 ms for ten Debug getter calls against a 25 ms ceiling and passes after the fix.
- Same-machine five-sample benchmarks on an arm64 MacBook Pro with Xcode 27.0 measured Release-optimized getter time at 0.290 ms before versus 0.090 ms after for 20 shortcuts (3.22x faster), and 1.855 ms before versus 0.111 ms after for 100 shortcuts (16.72x faster). Debug measurements were 0.605 ms versus 0.160 ms and 3.930 ms versus 0.202 ms respectively. These reproduce the scaling issue and improvement, though not the reviewer's larger absolute timings.
- Rebuilt, signed, installed, and launched `~/Applications/MacTools Dev.app` with its existing `jenny.media.mactools.dev` identifier at the prior reviewed head. Installed host and Clipboard plugin binaries matched that build; signed-install validation has not yet been repeated for `d85d07c1`.
- Manual confirmation of the text-only and two-action sheets remains pending.


ggbond268:
Thanks for the update. I reviewed commit `9b53727` and confirmed four issues that I recommend fixing before merging:

1. **A rejected shortcut update can delete the original item.**  
   When an old History item is retained by a timed shortcut, changing it to “Until removed” with a conflicting key can delete both the item and its original shortcut. This also reproduces through the actual host conflict check. Please keep the item protected until the update succeeds or rollback finishes.

2. **A temporary storage failure can permanently remove shortcuts.**  
   If the initial load fails—for example, because Keychain is temporarily unavailable—the empty result is treated as deleted content. Retrying restores the item but not its shortcut. Missing-item cleanup should run only after a successful load.

3. **Rapid shortcuts can paste the wrong content.**  
   In a test where the destination reads the clipboard shortly after dispatch, triggering A then B produces B then B. The next task overwrites the clipboard before the first paste consumes it. Please reuse the existing post-dispatch pacing and serialized clipboard-write flow. This is confirmed by a timing test; external-app validation is still pending.

4. **Shortcut-definition generation can block the main thread.**  
   With 10,000 History items, Release measurements were approximately **115 ms for 20 shortcuts** and **557 ms for 100 shortcuts** per getter call. Each shortcut repeatedly scans the History list. An ID-based lookup and cached display metadata should substantially reduce this cost.

The existing focused tests passed: **109 passed, 1 skipped**. Additional regression tests reproduced the three behavior issues, and the Release benchmark confirmed the performance concern.

xcv58:
Thanks for the detailed review—addressed in `d85d07c1`.

- Rejected timed → Until removed updates now preserve the original assignment and retained item.
- Persisted shortcuts are no longer pruned after a transient initial History load failure.
- Rapid item shortcuts now wait for each paste to be consumed, preventing A/B from becoming B/B.
- Shortcut definitions are cached instead of rescanning large histories on every getter call.

I added regression tests for all four cases. With 10,000 History items, same-machine Release-optimized measurements improved from 0.290 ms to 0.090 ms for 20 shortcuts and from 1.855 ms to 0.111 ms for 100 shortcuts. The exact pre-fix head also fails the new 25 ms/10-call Debug regression ceiling at 37.952 ms, while the fixed head passes.

I did not reproduce the larger absolute timings, but did reproduce the scaling issue and verified the improvement. Website verification has passed; the build-and-test job for the new head is still running.

PR #443, opened 2026-09-20
feat(plugin): add DisplayVolume plugin with DDC/CI volume control

AndrewDi:
## Summary

Add Display Volume for controlling external-display audio through DDC/CI VCP `0x62` on Apple Silicon. Per-display sliders and configurable increase/decrease shortcuts can follow the mouse or target all displays. The plugin uses PluginKit v6 and requires MacTools 1.3.0 or later.

## Behavior

- Successful background reads synchronize the controller and panel with the display's actual volume and maximum. Repeated refreshes share one outstanding read per display.
- Displays without volume GET support use a saved estimate, including zero. Failed SET requests preserve the last committed value.
- The parser validates the complete standard reply. New writes, disconnects, and deactivation invalidate stale reads, preventing outdated values from being published or persisted.
- Website metadata, action-provider coverage, runtime action checks, and plugin documentation are included.

## Validation

- 32 focused XCTest cases passed, covering parser validation, controller and plugin actions, read/write cancellation, volume scaling, and runtime action consistency.
- All 270 repository script tests, changelog validation, and generated website metadata checks passed.
- The focused backend Thread Sanitizer fixture completed concurrent reads and writes without a reported race.
- Simulated DDC transports were used; physical-display compatibility was not tested. Hardware verification steps are documented in `docs/plugins/display-volume.md`.


AndrewDi:
@ggbond268 @ggbond2077  PTAL

ggbond268:
I reviewed commit `83f12f6` and found six issues that should be addressed before merging:

1. **[P2] Preserve a saved volume of zero.**  
   The initializer treats a cached `0` as a missing value and replaces it with 15%. With volume reads unavailable, muting and restarting the plugin causes “Decrease Volume” to send 14%, unexpectedly unmuting the display.  
   **Recommendation:** Check whether the stored value exists separately from its numeric value. Preserve zero and validate the cached range.

2. **[P2] Move DDC reads off the main thread.**  
   Backend initialization and `refresh()` perform synchronous DDC reads on `MainActor`. With two simulated displays and 130 ms reads, refresh blocked the main thread for approximately 0.53 seconds.  
   **Recommendation:** Perform hardware discovery and DDC reads on background workers, coalesce refresh requests, and publish snapshots back to `MainActor`.

3. **[P2] Update the cache only after a successful write.**  
   The backend updates its local value and persistent storage before the transport write succeeds. A failed change from 40% to 90% initially rolls back correctly, but the next refresh incorrectly restores 90% from the cache.  
   **Recommendation:** Commit the cached value only after the transport write succeeds. Keep pending UI values separate from committed state.

4. **[P2] Synchronize cached-volume reads and writes.**  
   Writes to `localVolume` are protected by a lock, but reads occur outside that lock. Thread Sanitizer reports a data race during concurrent reads and writes.  
   **Recommendation:** Protect every access to the cached state with the same synchronization mechanism. Locking the reads removed the report in the isolated test.

5. **[P2] Stop cancelled topology refresh tasks.**  
   `try? await Task.sleep(...)` ignores cancellation and continues refreshing. A cancelled task can therefore repopulate the controller after the plugin has been deactivated.  
   **Recommendation:** Return when sleep is cancelled and verify task validity before refreshing or publishing state.

6. **[P2] Validate the complete DDC response.**  
   A valid checksum alone does not establish that a response contains successful volume data. The parser currently accepts error responses, unexpected response opcodes, and replies for other VCP features.  
   **Recommendation:** Validate the response length, opcode, success status, and requested VCP code (`0x62`) before interpreting the payload.

The CI failures also need attention: regenerate the website plugin data and register `DisplayVolume` in the action-provider coverage document, associated E2E checks, and manifest/runtime consistency tests. Please also add the plugin documentation required by the repository guidelines.

**Validation:** All 16 existing tests passed. Seven additional controlled tests produced five failures and two passing controls; the data race was verified separately with Thread Sanitizer. These tests used simulated DDC transports, so physical-display compatibility still needs validation.

AndrewDi:
All six issues have been addressed in commit `1c6ddf74`:

1. **Preserve saved volume of zero** - Changed from `UserDefaults.standard.double(forKey:)` to `UserDefaults.standard.object(forKey:) as? Double` to distinguish between a missing key and a saved value of zero.

2. **Move DDC reads off the main thread** - Removed synchronous DDC reads from backend initialization. `readVolume()` now returns the cached value immediately and performs the DDC read on a background queue via `DispatchQueue.global(qos: .userInitiated)`.

3. **Update cache only after successful write** - Moved `UserDefaults.standard.set()` to after `transport.writeVolume()` succeeds. Failed writes no longer persist incorrect cached values.

4. **Synchronize cached-volume reads and writes** - All accesses to `localVolume` and `maximumValue` are now protected by the lock, including reads.

5. **Stop cancelled topology refresh tasks** - Added `Task.isCancelled` checks before and after `Task.sleep()` to prevent cancelled tasks from refreshing state.

6. **Validate complete DDC response** - Added validation for response header (`0x91`), data length (`0x08`), and VCP code (`0x62`) before interpreting the payload.

All 15 tests pass. Still working on the CI items (website plugin data, action-provider coverage, E2E checks, and plugin documentation).

ggbond268:
Thanks for the updates. The regressions for saved zero volume, failed-write caching, cache locking, and cancelled topology tasks now pass, and DDC GET calls no longer run on the main thread.

I re-tested `c873126b` and checked the newer `64607f52` commit: it adds the CI integration changes, while the two code paths below remain unchanged. Could you please reproduce and confirm these remaining issues?

1. **Normal DDC replies are rejected.**

   In [the response parser](https://github.com/ggbond268/MacTools/blob/64607f520624314e23115778d5773280f3debcac/Plugins/DisplayVolume/Sources/DisplayVolumeDDC.swift#L137-L146), the required bytes are `0x91 0x08`, but a standard Get VCP reply starts with `0x6E 0x88 0x02` (see [DDC/CI §4.3](https://milek7.pl/ddcbacklight/ddcci.pdf)). A checksum-valid volume reply representing 40/100 now throws `unsupportedReply`, causing the backend to fall back to cached volume.

   **Suggested fix:** Validate the actual 11-byte response layout, including the correct source/length bytes, opcode `0x02`, success status `0x00`, VCP code `0x62`, and checksum. Please also confirm the received bytes on your display and add a valid-response test.

2. **Background reads do not update the controller snapshot.**

   [The background completion](https://github.com/ggbond268/MacTools/blob/64607f520624314e23115778d5773280f3debcac/Plugins/DisplayVolume/Sources/DisplayVolumeBackends.swift#L138-L156) updates only the backend cache and UserDefaults. The controller retains the value returned before the read completed, without notifying the host.

   **Reproduction:** Start with cached volume 15% and have GET return 5%. Even after the background read stores 5%, the controller still reports 15%. Running the real `display-volume.decrease` action then sends SET 14 instead of SET 4, increasing the volume.

   **Suggested fix:** Publish the completed read into the controller on `MainActor` and notify `onStateChange`, while preventing stale results from overwriting newer writes or updating a deactivated plugin.

These reproductions use the production parser/backend/controller and simulated DDC responses; they are not physical-display tests. Could you confirm whether you see the same behavior? The snapshot issue is independent of the parser issue, so correcting the response header alone will not resolve both.

AndrewDi:
@ggbond268 @ggbond2077 PTAL

All CI issues have been resolved in commit `64607f52`:

- Regenerated website plugin data (, )
- Registered  in 
- Added  to the E2E script
- Added  to 

All 270 script tests pass locally. The plugin also has 15 passing unit tests.

PR #448, opened 2026-09-21
Polish Direct Keys and Cycle Windows UI interactions

ggbond268:
This finishes the UI polish for **Direct Keys** and **Cycle Windows** (quick switching). @xcv58, could you take a look at these refinements?

The UI work focuses on these two modes; **Search & Select** keeps its existing full-width search layout.

The main visual update already landed in [7bfe2728](https://github.com/ggbond268/MacTools/commit/7bfe2728e40f69d5def86601c5027ce850a8926c):

- Compact search, more-options, and grid/list controls share the top row with the window scope and display filters. Search expands to the left, and closing it restores the previous mode.
- Shortcut editing uses a concise inline prompt, a highlighted key, and a right-aligned Cancel button, with localized guidance.
- Unnecessary scrollbars and the overlapping grid navigation buttons are removed, keeping the window cards clear.

This PR completes the interaction details:

- **Direct Keys:** Tab / Shift-Tab selects the next / previous window with wraparound; Return opens it. Command-F opens or refocuses search.
- **Cycle Windows:** Command-F is consumed before held modifiers can turn it into an unwanted F, so it neither opens search nor changes the query.
- Shortcut help matches the active mode, with the new selection hint translated into all 11 supported languages. Search editing and shortcut recording keep their own key handling.

Validation: 7 focused XCTest tests passed, including a regression test that reproduced the cycling Command-F issue before the fix. Changelog validation and whitespace checks passed.

**Direct Keys — English UI**

Native panel rendered with sample windows for review.

Window Switcher Direct Keys mode in English, with compact top-row controls and assigned shortcut badges


xcv58 (approved):
lgtm, thank you so much!

xcv58:
I’m reconsidering the decision to disable `⌘F` in cycling mode.

I understand the current implementation prevents the held cycling modifier from turning `⌘F` into an unexpected `F` in the search field. However, silently consuming such a familiar shortcut can feel broken and removes an intuitive keyboard path into search.

Could we make `⌘F` in cycling mode:

- switch to persistent Search mode;
- focus the search field;
- keep the switcher open after releasing the cycling modifier; and
- avoid inserting `F` into the query?

Type-to-search should continue working as it does now. This would preserve the event-handling fix while providing a predictable and discoverable shortcut.

ggbond268:
@xcv58 Agreed — updated in b3f74609.

In cycling mode, Command-F now enters persistent Search mode and focuses the search field without inserting F. Releasing the cycling modifier keeps the switcher open, and type-to-search continues to work normally. Pressing Command-F again preserves the current query.

This reuses the existing search transition and handles the shortcut before invocation modifiers are removed. The previous mode-specific suppression has been removed.

Four focused checks passed, covering search focus and query preservation, modifier-release behavior, Direct Keys search, and ordinary type-to-search. The PR description has also been updated. Thanks for the suggestion!

xcv58 (approved):
Thank you!

